PR #35, opened 2014-03-27
Add noGui to the list of configurable options

haasn:
So I don't have to alias or manually type --no-gui when invoking


Slider-Whistle:
Please reconsider, I hate having to use things like bash scripts or aliases. I generally agree that launch arguments shouldn't permanently set configuration options permanently, but I don't see anything about the PR that really implies that. Is that just how Syncplay works? I think it'd be better to change that.

I don't like to have many different windows open at once, and I think this is definitely one of those kinds of programs that shouldn't need any more than one. I never run mpv in anything that's not fullscreen either, so having to minimise it to do things is pretty inconvenient.

PR #49, opened 2014-12-01
German Translation finish

DerGenaue:
Finished translation although one might find better words at some points
- German Text should now be free of any typos (I let a spell-checker run)
- corrected one typo in English text


Et0h:
Syncplay 1.3.0 Beta 2 comes with new features (update checker, user readiness) and other tweaks ('controlled rooms' and 'room controllers' are now 'managed rooms' and 'room operators'), so there are a few more translation tasks.

PR #63, opened 2015-05-27
Stop using --slave-broken

wiiaboo:
mpv finally dropped --slave-broken, so Syncplay on Windows needs this change to work.


Et0h:
Cheers. I'll test it when I get the latest lachs0r build that takes into account https://github.com/mpv-player/mpv/commit/e8a1d35e9c439ba8a86507e02292cc24fef0174a

PR #69, opened 2015-08-03
Don't overwrite user arguments with our style defaults

wiiaboo:
This allows the per-player user arguments to overwrite
Syncplay's style defaults in mpv.

Be careful that, if you turn off force-window and idle, you have to
supply Syncplay with the filename before opening the video window.


Et0h:
Thanks for your contribution, wiiaboo :)

artur309:
nice

PR #71, opened 2015-08-14
Update messages.py

BrokeRU:
Translation to Russian.


Et0h:
Yay, another Russian translation update. Thank you very much, BrokeRU! :+1:

PR #76, opened 2015-10-06
German Translation

DerGenaue:
This makes better sense

### DIFF
--- a/buildPy2exe.py
+++ b/buildPy2exe.py
@@ -102,14 +102,14 @@ NSIS_SCRIPT_TEMPLATE = r"""
   LangString ^UninstConfig $${LANG_RUSSIAN} "Удалить файл настроек."
 
   LangString ^SyncplayLanguage $${LANG_GERMAN} "de"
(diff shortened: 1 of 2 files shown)


nilsding:
This translation seems accurate.  The "comment" in `build2exe.py` on line 105 is imho unnecessary (and IIRC not a valid NSIS comment).

Et0h:
Yes, the proposed comment on line 105 results in the following error:
>Traceback (most recent call last):
>   File "/syncplay/buildPy2exe.py", line 608, in compile
>    raise RuntimeError("NSIS compilation return code: %d" % retcode)
>RuntimeError: NSIS compilation return code: 1

DerGenaue:
Oops, now I get it;
My fault; I'm used to work with Java and other Languages and I keep mixing up the syntax.
The editor highlights it => I'm done here

DerGenaue:
By the way: what's IIRC and NSIS?

Et0h:
Ah, sorry for not being more clear. 'IIRC' means 'If I Recall/Remember Correctly' and NSIS means 'Nullsoft Scriptable Install System' (i.e. what Syncplay uses for making Windows installers).

Issue #73, opened 2015-09-07
High-latency player connections can cause "fighting"


tari:
I've observed this behavior when working on Kodi/XBMC support, and I suspect #52 is related since the symptom ("it actually does manage to pause the video, but the madVR user will start it immediately again") is the same.

My implementation of control of Kodi involves connecting over HTTP, making a request and waiting for a response. Due to how the API is designed, most actions require two round-trips (a state query to get a necessary parameter value, then the actual command/query). As I've implemented it now, every query opens a new HTTP connection. The upshot of all this is that operations on a Kodi player tend to have high latency.

Where things start going wrong is when another user in the room performs an action on playback state (play, pause, seek). A state change arrives from the server and a command is dispatched to the player, but a preceding status check will usually complete first. This causes a conflicting state to be reported, and latency effects tend to make the player cycle between playing and paused states.

(Performing blocking RPCs might be an easy fix but would tend to bog down the reactor a lot, so it's not a good solution.)

Representative debug output:

```
[11:17:07] client/server << {"State": {"ignoringOnTheFly": {"server": 1}, "playstate": {"paused": true, "position": 2
6.0, "setBy": "None", "doSeek": null}, "ping": {"serverRtt": 0.1529710292816162, "clientLatencyCalculation": 1441646227.429288, "latencyCalculation": 1441646227.505084}}}
[11:17:07] <None> paused
[11:17:07] client/server >> {"State": {"ignoringOnTheFly": {"server": 1}, "ping": {"clientRtt": 0.1514420509338379, "clientLatencyCalculation": 1441646227.581642, "latencyCalculation": 1441646227.505084}, "playstate": {"paused": false, "position": 27.008032083511353}}}
[11:17:07] client/server << {"State": {"ignoringOnTheFly": {"server": 1}, "playstate": {"paused": false, "position":
27.082831205780106, "setBy": "None", "doSeek": null}, "ping": {"serverRtt": 0.1451740264892578, "clientLatencyCalculation": 1441646227.581754, "latencyCalculation": 1441646227.65037}}}
[11:17:07] <None> unpaused
[11:17:07] client/server >> {"State": {"ignoringOnTheFly": {"server": 1}, "ping": {"clientRtt": 0.14456892013549805,
"clientLatencyCalculation": 1441646227.727131, "latencyCalculation": 1441646227.65037}, "playstate": {"paused": false, "position": 27.05450201034546}}}
[...]

Issue #52, opened 2015-01-31
madVR Support


Ooglogput:
As it stands, if you use madVR, other people can play and skip just fine, but they can't pause. I'm pretty sure this is a known issue, but formal reports are always good.

Et0h:
Yes, it is a known issue and is noted on the troubleshooting page available at http://syncplay.pl/guide/trouble/ ("If you are using MPC-HC and have set the video filter to madVR then you will need to revert to the default filter. madVR is currently incompatible with Syncplay.")

As the issue has yet to be resolved there's nothing wrong with prompting me to give it some further consideration. User reports are a valuable tool to help prioritise the use of limited development resources.

 I seem to recall that it was madVR either not listening properly to the MPC-HC API or giving it incorrect information and that we can't make madVR work with Syncplay from our side. However, I will look into:
* The precise cause of the issue, and whether it can be overcome via a workaround within Syncplay;
* Whether this issue has been reported to the relevant MPC-HC/madVR developers, and whether a fix has already been issued;
* Whether MPC-HC can be told not to use madVR with Syncplay (e.g. through a command-line argument), and whether this would cause more problems than it solves; and
*  If it is not fixed from Syncplay's side, making it clearer on the Syncplay website that madVR is not supported (or is only supported in newer version of the related software if it gets fixed), and if possible detecting madVR messing things up so Syncplay can give an appropriate warning message.

Ooglogput:
You can pause the video with through your own Syncplay client though; does that use a different function from when someone else wants to pause?

schoerg:
madVR has a network interface, from which you can control the clients playback, described in "net-protocol.txt". If the bug is not fixed you could look at this.

Et0h:
**Ooglogput**: I'll look into the relevant behaviour to see if I can obtain a better understanding of what is happening. It was a long time ago that it was last investigated.

**schoerg**: Thanks for the heads up. It could potentially be very useful, although hopefully it won't be necessary to support yet another API (which could result in its own compatibility issues and would need maintaining).

Ooglogput:
Might also be worth mentioning that MPC-HC's own web interface does work with madVR; at least for everything I've tested.

Et0h:
I just tried and wasn't able to replicate. Could you please let me know if the problem occurs with the latest version of madVR and MPC-HC. If so, do you have any other changes from the default, e.g. a different subtitle renderer, and if that is disabled does that make a difference?

[...]

PR #77, opened 2015-10-16
Add player command latency compensation

tari:
Somewhat experimental fix for #73. Worked very well with XBMC/Kodi (coming soon to a PR near you), didn't seem to have any ill effect on VLC or MPC, the two existing players I tested with.

`playerCommandDelay` should generally be set to an amount of time longer than the expected largest delay between sending a play/pause/seek command and having that command reflected in the status reported back to the client. Some kind of autoconfiguration (or just a warning if "fighting" is observed) would be nice, but this at least provides a configuration option for user adjustment.

With appropriate configuration, I suspect this fixes issues with madVR when "delay playback start until render queue is full" as well, as reported in #52.

### DIFF
--- a/syncplay/client.py
+++ b/syncplay/client.py
@@ -175,10 +175,15 @@ class SyncplayClient(object):
         seeked = _playerDiff > constants.SEEK_THRESHOLD and _globalDiff > constants.SEEK_THRESHOLD
         return pauseChange, seeked
 
+    def _ignoringPlayerStatus(self, cookie=None):
+        if cookie is None:
+            cookie = time.time()
+        return cookie < self._lastPlayerCommand + self._config['playerCommandDelay']
+
     def updatePlayerStatus(self, paused, position, cookie=None):
         # Ignore status report if the cookie is stale
-        if cookie is not None and \
-                cookie < self._lastPlayerCommand + self._config['playerCommandDelay']:
+        if self._ignoringPlayerStatus(cookie):
+            self.ui.showDebugMessage('Ignoring stale player status with cookie {}'.format(cookie))
             return
 
         position -= self.getUserOffset()
@@ -515,6 +520,7 @@ class SyncplayClient(object):
     def setPosition(self, position):
         if self._lastPlayerUpdate:
             self._lastPlayerUpdate = time.time()
+        self._playerPosition = position
         position += self.getUserOffset()
         if self._player and self.userlist.currentUser.file:
             if position < 0:
--- a/syncplay/client.py
+++ b/syncplay/client.py
@@ -258,6 +258,10 @@ class SyncplayClient(object):
 
     def _serverUnpaused(self, setBy):
         hideFromOSD = not constants.SHOW_SAME_ROOM_OSD
+        # In high-player-latency situations we might report our state back to
+        # the server before any player status is accepted as fresh. Override
+        # the locally-stored playback state.
+        self._playerPaused = False
         self._playerCommand(self._player.setPaused, False)
         madeChangeOnPlayer = True
         self.ui.showMessage(getMessage("unpause-notification").format(setBy), hideFromOSD)
@@ -265,6 +269,7 @@ class SyncplayClient(object):
 
     def _serverPaused(self, setBy):
         hideFromOSD = not constants.SHOW_SAME_ROOM_OSD
+        self._playerPaused = True
         if constants.SYNC_ON_PAUSE and self.getUsername() <> setBy:
             self.setPosition(self.getGlobalPosition())
         self._playerCommand(self._player.setPaused, True)
--- a/syncplay/client.py
+++ b/syncplay/client.py
@@ -110,6 +110,8 @@ class SyncplayClient(object):
         self.autoPlay = False
         self.autoPlayThreshold = None
 
+        self._lastPlayerCommand = time.time()
+
         self.autoplayTimer = task.LoopingCall(self.autoplayCountdown)
         self.autoplayTimeLeft = constants.AUTOPLAY_DELAY
 
@@ -137,6 +139,16 @@ class SyncplayClient(object):
             constants.OSD_WARNING_MESSAGE_DURATION = constants.NO_SECONDARY_OSD_WARNING_DURATION
         self.scheduleAskPlayer()
 
+    def _playerRequest(self, f, *args, **kwargs):
+        """Send a request with cookie to the player."""
+        kwargs['cookie'] = time.time()
+        f(*args, **kwargs)
+
+    def _playerCommand(self, f, *args, **kwargs):
+        """Send a command to the player, affecting cookie freshness."""
+        self._lastPlayerCommand = time.time()
+        f(*args, **kwargs)
+
     def scheduleAskPlayer(self, when=constants.PLAYER_ASK_DELAY):
         self._askPlayerTimer = task.LoopingCall(self.askPlayer)
         self._askPlayerTimer.start(when)
@@ -145,7 +157,7 @@ class SyncplayClient(object):
         if not self._running:
             return
         if self._player:
-            self._player.askForStatus()
+            self._playerRequest(self._player.askForStatus)
         self.checkIfConnected()
 
     def checkIfConnected(self):
@@ -163,14 +175,19 @@ class SyncplayClient(object):
         seeked = _playerDiff > constants.SEEK_THRESHOLD and _globalDiff > constants.SEEK_THRESHOLD
         return pauseChange, seeked
 
-    def updatePlayerStatus(self, paused, position):
+    def updatePlayerStatus(self, paused, position, cookie=None):
+        # Ignore status report if the cookie is stale
+        if cookie is not None and \
+                cookie < constants.PLAYER_COMMAND_DELAY + self._lastPlayerCommand:
+            return
+
         position -= self.getUserOffset()
         pauseChange, seeked = self._determinePlayerStateChange(paused, position)
         self._playerPosition = position
         self._playerPaused = paused
         if pauseChange and utils.meetsMinVersion(self.serverVersion, constants.USER_READY_MIN_VERSION):
             if not self.userlist.currentUser.canControl():
-                self._player.setPaused(self._globalPaused)
+                self._playerCommand(self._player.setPaused, self._globalPaused)
                 self.toggleReady(manuallyInitiated=True)
                 self._playerPaused = self._globalPaused
                 pauseChange = False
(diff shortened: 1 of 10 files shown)


Et0h:
If Syncplay detects the relevant change has actually been implemented (e.g. because the player has a play/pause or position state consistent with the most recent desired change), does this override playerCommandDelay time, or would Syncplay continue to ignore the player even though it had indicated that it had implemented the relevant command?

tari:
Right now it continues to ignore the player, yes. It could watch for the "expected" state, but since the delay is honored for seeks as well doing so could be tricky.

The benefit of skipping the delay, as I see it, is faster reaction to some state changes. I'm not convinced it's worth the complexity, since sufficiently long command delays will be compensated for by seeking if necessary- skipping the delay just eliminates some seeks.

Uriziel:
Ok, I gave it some thought.
For a second lets discard all changes in player classes and `_playerRequest` method and cookie idea, but keep `_playerCommand`.

Wouldn't following achieve the same result if executed there (not tested)?

    self._askPlayerTimer.stop()
    self._askPlayerTimer.start(constants.PLAYER_ASK_DELAY, false)   
    # Reschedule ask player to time.time() + PLAYER_ASK_DELAY 
    # so players can update it's state after receiving a command

Also, I don't like idea of making cookie value configurable (it's too hard to comprehend it's function to modify)

tari:
> Wouldn't following achieve the same result if executed there (not tested)?

It doesn't work in the case where a call to `_playerCommand` happens between a call to `askForStatus` and the player calling back to `updatePlayerStatus`. Without a cookie to flag that the data coming into `updatePlayerStatus` is stale, the client could wrongly see the old status.

> Also, I don't like idea of making cookie value configurable (it's too hard to comprehend it's function to modify)

I think it's important to be configurable, since selection of an appropriate value could depend on many variables (machine speed, network conditions..) and longer delays, while not exactly harmful, tend to be more annoying (causing more seeking on pause).

In general, I'd say the guideline is "increase this value if you see rapid switches between playing and paused without any input, but keep it as low as possible " which is pretty easy to understand. There's probably room for UI improvement to make that more obvious, but I don't think removing the option is the right answer.

Uriziel:
Okay. It seems you're right on that one.

Now the question is: what are the consequences of setting playerCommandDelay on a high (or any) value, I don't think I figured them all yet. 
Let's say I set it for 5 seconds, what's then?

tari:
I'm speculating a little here since I haven't tested it, but mainly it'll act as if the player is stalled while waiting for the timeout to expire.

For example, a room has two users A and B. B has a 5-second command delay. Both players are initially paused. On unpause, A begins playing with negligible latency while B's command delay causes the client to report as playing, but time is not advancing. This quickly becomes a situation where B appears to be lagging, so A rewinds to be back in sync. When B's delay expires, their client reports the actual player time, which is up to 5 seconds ahead of what B's client was reporting. This causes more seeking.

In the worst case, B rewinds to be back in sync with A at this point, and once again is subject to its command delay which could put both clients in a loop of seeking. Otherwise A can seek forward to be back in sync, which isn't an issue.

---

So I guess this implementation trades play/pause loops with seek loops in some configurations, but that could be fixed by artificially advancing the client's time in a fashion similar to how it currently records the playing/paused state on sending a command to the player.

Uriziel:
Check out my edit on previous post.
I think loop / rewind should not occur because of `self._lastPlayerUpdate` but I might be wrong and it needs testing

tari:
Real-world data point: I need about 150ms delay for robust sync across my LAN with Kodi (with syncplay controlling Kodi on another machine).

I tested some with two clients, each with a 5-second delay and observed no looping. Same result if one has a long delay but the other none, so it looks we're okay on that count.

---

The question then is how to handle the option controlling this functionality. What it looks like now, in the GUI:

screenshot

The short description isn't very meaningful to most users, whereas I think the tooltip explains when you might want to change it (though the current wording could be improved). I could leave it as is but improve the tooltip, add better explanation to the label (so users don't need to read the tooltip to learn what it's good for), or do something else that sounds better to everybody.

Uriziel:
On the other hand I tested it out got heavy loops after seeking.
Cookie introduces following issues: 
- Executing two player commands at once:
    - If you pause and unpause (i.e. miss-click) your player actually starts playing, but everyone's else will start only after the delay (introducing desync and slowdowns).
   - If you seek and unpause (common for i.e. "Did you see that") it yet again introduces delay.
- When your player was seeked by some other client right before updating it's global status it literally breaks. It's fun issue. 
  - Your player seeks and starts ignoring incoming statuses for the delay.
  - Your player reports its position based on `_lastPlayerUpdate` thus it reports it's position from before the seek.
  - People are rewinded / seeked to old position.
  - On a good day your new position is reported to Syncplay before other clients make you seek back. You execute seek yet again.
  - On a very good day everything breaks actually.
Setting delay value low though mitigates those issues. At `0.05` it should be actually unnoticeable.

I think your way of thinking is correct and it addresses some issues. I'm thankful for your insight. Yet it might require some modifications or getting different approach.

Readjustment of delay value could be based on player / set by the player after approximating necessary value (i.e. calculated by your delay with Kodi). 

Cookie might be actually not necessary for mpv at all (I think it handles API calls on separate thread thus is very fast in replying), but apparent in MPC (all API calls are delegated to main thread and can be delayed if it's stuck processing something else).

I think rescheduling askForStatus (as provided in previous example) might be a nice start too. It should at least fix half of the issues (askForStatus is executing in the main twisted loop). Forcing cookie to be no longer than `PLAYER_ASK_DELAY` is another possible improvement.

Other than that, it's possible to make askForStatus -> updateLocalState commands async (i.e. deferred implementation) and all the delays could be handled within player classes. Which I think is good idea to consider, but haven't thought it through.

@tarl Please express your thoughts on this and let's leave the decision for @Et0h

tari:
> Executing two player commands at once: 

Yeah, that's an inherent limitation of this approach. The delay should be kept as low as possible for any given configuration.

> Your player reports its position based on _lastPlayerUpdate thus it reports it's position from before the seek.

This can be improved or fixed by handling seeks in the same way play/pause commands are (update the client state on sending the command, assuming that's what the player will do).

> Setting delay value low though mitigates those issues. At 0.05 it should be actually unnoticeable.

We could even make the default delay 0, which would ensure none of the new (potential) issues can occur in the default configuration. Similarly, instruct users to increase the value only if necessary.

---

Considering that this patch (or more generally handling slow players) blocks XBMC/Kodi support, here's what I propose for merging:

* Change the default command delay to 0, so users who don't change that will see no change in behavior.
* Tweak the GUI for this option to discourage changing it unless actually necessary, clarify its purpose some.

With that machinery in place, we ensure it won't break existing users but it provides a platform to build on. I can then land the new player module I've been working on, and work can go into improvements such as:

* Investigate alternate approaches to the problem
* Autoconfiguring command delay
* Rescheduling `askForStatus` (saving some wasted work during delays and generally improving player status latency by 0.5*`PLAYER_ASK_TIME` following a command)

Et0h:
I think many of the issues could be resolved if Syncplay knew what the player state should be assuming recent changes had yet to be executed, and what the state should be assuming all changes had been implemented.

With this information Syncplay can have a good sense of whether the player had successfully implemented all requested changes so that it could start trusting what the player said again, and in the meantime report to the server a best-guess situation based on what the position and play/pause state should be assuming the state changes had been implemented (and, if playing, assuming that position will move forward at 1 second per second).

Some of this work already exists for VLC and mpv (e.g. with getCalculatedPosition) but may need more testing.

tari:
That should be pretty easy. It's already doing that for `_playerPaused` in `_serverPaused` and `_serverUnpaused`, and we already do an adjustment based on the last player update time in `getPlayerPosition`, so updating `_playerPosition` on server seek will probably do the job.

tari:
I hammered on 094a204 with a lot of seeking and couldn't get it to do anything bad; seems okay?

Et0h:
Re: Line 523 of client.py, not sure if it one is supposed to set self._playerPosition to position before or after the user offset has been applied - will need to check.

Et0h:
Tested things out and things seem to work okay with low ignore levels set. With high ignore levels there can be some lag if you seek just after someone else, or if you pause/unpause and seek.

Et0h:
One possibility is that the seconds to ignore is set as "Auto" in settings GUI by default, and this then uses a per-player default setting. Then people can specify a different number if they wish to deviate from the per-player default. This can be achieved in the GUI by reducing minimum value of the spinbox and using setSpecialValueText.

tari:
The problem with an Auto setting is that a user can't tell what that automatic value is- if the default isn't high enough, they'd have to guess rather than having a starting point to work from.

I suppose the configuration option could be an adjustment from a per-player default, so the initial value is always zero and the actual amount of time involved depends on the player in use. Then the user's approach is simply to give it a positive value if the delay is not long enough in their configuration, or a negative value if they want to make the delay shorter.

This also helps prevent some problems where somebody might have a custom delay configured for a particularly slow player, but they switch to another one. If the options were "auto" or a fully-specified time delay, the faster player would tend to have a much longer delay than required if the user does not also change the delay setting. An adjustment from the default may still have a longer delay than necessary for the new player, but the actual length of that delay will tend to be lower.

Et0h:
Hmm, I see what you mean. If the latency issue only affects MPC-HC and XBMC then I guess one option would be to allow for per-player configuration of the delay values for these two players.

tari:
That's an option, but would require more work to support in the GUI (AFAICT the configuration GUI doesn't do anything on a per-player basis so it would need some new machinery to do so).

Or I guess the easy choice is to duplicate the option for each player that might need it, so with the current design (for example) there would be two spinboxes, one each for MPC-HC and XBMC.

Et0h:
Syncplay stores custom player command line arguments on a per-player-path
basis. However, I want to avoid clutter of the main tab. Could reuse the
way data is stored for that but use player names rather than paths, and
have it in misc. So long as it is stored in a flexible sensible
player-agnostic manner we can start with two spin boxes and have the option
to change it later without the config file getting out of date.

Uriziel:
@tari: I merged this PR to shared-playlists branch as we test that thoroughly ATM. Though it had some conflicts so you might check into that.

If there aren't any issues found it will be just merged.

Et0h:
We tried this out in the Shared Playlist branch and it caused issues. Trying to do two actions in quick succession from the client was problematic, and it made Syncplay feel much less responsive.

The current view is that ignoring the player after a command has been issued before it has been implemented should be handled by the player code, as that can best take account of the behaviour of the specific player.

Uriziel:
Slight clarification:
Player code = Syncplay's player class

PR #78, opened 2015-10-23
Probe player paths in a non-UI thread

tari:
Move calls to `getPlayerIconByPath` into a background thread so the GUI doesn't lock up when those calls take a long time.

This is particularly important for supporting remote players which are specified by URL, like Kodi. Determining if a player exists at the given path requires hitting the network, and invalid paths often require a timeout to determine they are indeed invalid. It may also be useful with slow filesystems.

While the thread is working, the player icon is replaced with an animated spinner to indicate it's busy: 
spinner. (The actual version looks better than this gif since it has transparency.)


Et0h:
Thanks for this. What's the copyright license on the .mng file and who was the author?

tari:
I created it myself with the tool [here](http://spiffygif.com/) which seems to have disappeared in the past few days. From an archived version, its output can be used freely:

> * You are free to copy and redistribute any images generated from SpiffyGif in any medium or format, even commercially.
> * You are permitted to remix, transform, and build upon images generated from SpiffyGif for any purpose, even commercially.

Et0h:
Oh, excellent :)

Et0h:
I note that the spinner.mng resource has not been added to the list of resources to be included in BuildPy2exe.py. If Uriziel is happy with the PR then this would be an easy thing to fix, though.

PR #79, opened 2015-10-26
Ignore Spaces, etc... from host input

DerGenaue:
Request by nnamrrehdlopoel who likes to Copy- / Paste Spaces

### DIFF
--- a/syncplay/ui/GuiConfiguration.py
+++ b/syncplay/ui/GuiConfiguration.py
@@ -291,6 +291,7 @@ class ConfigDialog(QtGui.QDialog):
         self.processWidget(self, lambda w: self.saveValues(w))
         if self.hostCombobox.currentText():
             self.config['host'] = self.hostCombobox.currentText() if ":" in self.hostCombobox.currentText() else self.hostCombobox.currentText() + ":" + unicode(constants.DEFAULT_PORT)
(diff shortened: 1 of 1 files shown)


Et0h:
Nice idea, although I tried the code on PyCharm/Windows 7 and it didn't work

The error I got was:
>Traceback (most recent call last):
>  File "\syncplay\ui\GuiConfiguration.py", line 294, in _saveDataAndLeave
>    self.config['host'] = self.config['host'].translate(None, " \t\n\r")
>TypeError: translate() takes exactly one argument (2 given)

My understanding is that this is because translate() behaves differently depending on whether the input is unicode or non-unicode.

Things worked with the slightly less elegant code:
> self.config['host'] = self.config['host'].replace(" ","").replace("\t", "").replace("\n","").replace("\r","")

Things also worked if I added a conversion to utf-8, although I'm not sure if that may also create problems:
>self.config['host'] = self.config['host'].encode('utf-8').translate(None, " \t\n\r")

DerGenaue:
Actually I just googled it and used the first solution I found ;)
It doesn't matter how it is done as long as it works ;)
Thx for testing and correcting

Et0h:
Code is now in sharedplaylist branch 18a17b0aa96c6eae11862740b94421a433a24a93

PR #85, opened 2015-12-14
change type, of file in ConfigurationGetter.py from str to utf-8

abhsag24:
changed the type, of file  in argparser so that files can accept unicode characters


Et0h:
Tested in Windows and it doesn't seem to break anything. Unicode for file path arguments in Windows is already broken (due to how Python gets the arguments), so people have to open Syncplay then the file if they want to open Unicode files.

Issue #81, opened 2015-12-07
Still need help with developing for OS X?


kylestev:
I'd love to help!

kylestev:
After downloading it and running `make` it would appear so... I'll poke around some more in the morning. It's getting late where I live.

nilsding:
Running `make` fails because the file system layout on OS X unlike to those on other Unix-like systems.  Also, I think it would be better to package Syncplay as an .app on that system, rather than copying files to where they belong.

kylestev:
changing it to use /usr/local as the prefix seems to work as /usr is owned
by root. /usr/local is used by homebrew so it's not an uncommon pattern as
far as I know to use that directory. packaging it as an app is definitely
something to consider. the issue with that is you'll run into problems with
OS X's security settings if you aren't a verified publisher. I'm not quite
sure what the workflow is for that if you want to distribute the app
outside the App Store.

nilsding:
It seems that most people usually create a disk image (i.e. `.dmg`) big enough to hold the `.app` directory and a symlink to the `/Applications` directory.  Regarding OS X's security settings, depending on your setting, it will display [a warning](https://cloud.githubusercontent.com/assets/1809170/11623724/0de04b28-9ccf-11e5-924c-bbbaa5a10da1.png) on the first start and you have to open the System Preferences app to [allow the system to run it](https://cloud.githubusercontent.com/assets/1809170/11623695/ea0f393e-9cce-11e5-8462-90d278650fa4.png).

Also, for packaging Python applications as an `.app` there is a tool/library/whatever similar to `py2exe`, namely [`py2app`](http://pythonhosted.org/py2app/).  Let's see if this thing works.

Et0h:
Yes, we certainly still need help getting Syncplay to work properly on OS X. The ones that I am aware of are:

**1.  A way to generate a native package for Mac that "just works".**
I don't know the terminology (.app files?), but if we can make some sort of installer which people can use to run Syncplay on Mac without Miniconda, command line, etc then that'd be great. Ideally it would be easily generated on a Windows or Linux machine, so we can make a new package/installer whenever a new version of Syncplay is out.

**2. Double check installation instructions**
If the above is not possible, review the instructions at http://syncplay.pl/guide/install/#OSX and suggest any changes if they are necessary.

**3. Fix the menu bar in OS X**
At present the  PySide menu bar works on Windows and Linux, but not OS X. Possibly running it natively will fix that, but whatever the case we want it to show up.

[...]

PR #86, opened 2015-12-19
bugfix: display the menu bar in OS X

nilsding:
A picture (or screenshot) says more than a thousand words:

screen shot 2015-12-19 at 19 41 11

Now let's work on the other issues @Et0h [mentioned](https://github.com/Syncplay/syncplay/issues/81#issuecomment-162468488) in #81.


Et0h:
Yaaay! :+1:

blaenk:
Great work :smile:

It looks like in order to change the application name from "python2" to something else, we need to edit some `Info.plist` file in the generated .app (if we do generate one, not sure about that). [Source](http://stackoverflow.com/questions/7827430/setting-mac-osx-application-menu-menu-bar-item-to-other-than-python-in-my-pyth)

Issue #80, opened 2015-12-06
Syncplay Server should tell the IP and port of the machine its running on.


abhsag24:
right now after starting the server, the user has to search their own ip and port, to give to the clients, it should be the responsibility of the syncplay server to tell the user their ip and port.

Uriziel:
Not really possible, I guess? Server does not necessarily knows its IP, nor it's bound to any particular interface so it's accessible from multiple ips (like localhost)

abhsag24:
the server could provide a list of all ip's its listening on? i mean if there are many interfaces it could provide the ip's of all of them?

Uriziel:
Still there's no easy way to tell (platform dependent too, I guess?). Also its ip might be hidden by NAT or something similar

Et0h:
Port is easy. We could add an option to connect to syncplay.pl to get Internet-facing IP and host a special script, but people could just do wget -O - -q http://whatismyip.org/ or whatever to get their IP.

abhsag24:
People Can find their own IP's, but shouldn't syncplay do it for them? not everybody is a power user? , why not have syncplay return the output from  wget -O - -q http://whatismyip.org/  ?

Et0h:
We wouldn't want to use any old random service as at any point they could go down or change their response. Furthermore, it might not be a good idea to connect to an external IP without user's consent for privacy reasons. So if we did it we'd need to host our own script on our own server so we could be sure it'd work (so long as our server is up). If someone is happy to ocnnect to Syncplay server, we could also check to ensure server is running latest stable version of Syncplay (or newer). I don't see it as a top priority, though, as non-power-users can simply use the public Syncplay server.

abhsag24:
we could send have a socket connect connect to http://syncplay.pl ,use the getsockname function to know which address the connection was established with . For example:

```python
import socket
s = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
s.connect(("syncplay.pl",80))
print(s.getsockname())
s.close()```

Et0h:
Does getsockname work properly across all platforms (operating systems) and circumstances (network configurations)? Does it have any advantages over connecting to a custom PHP script via urllib which also checks for updates?

abhsag24:
[...]

PR #88, opened 2015-12-23
Show servers own ip address and port

abhsag24:
displays local ip's by seeing all interfaces and excluding loopback, and non consequential ip's(that start with, 169.254)
displays the ip of network by, connecting to syncplay.pl, and then gethostname, will change to twisted way, if it is important.


Et0h:
Thanks for your contribution. Does this connect to syncplay.pl without asking for the user's permission or giving them an opt-out? If so, I'm not happy about that element of it for the reasons stated at #80 (i.e. that "it might not be a good idea to connect to an external IP without user's consent for privacy reasons")

Et0h:
I also note that it would also ideally want any messages moved to messages.py with appropriate stubs so as to be compatible with Syncplay's support for languages other than English.

abhsag24:
At, the moment it does'nt give an opt-out, i'll move the messages in the next commit, what do you propose should be the ideal way of having an opt-out through raw_input, or the arg_parser?

Et0h:
In relation to consent I think an opt-in would be better. In terms of how, Syncplay already uses argparser for command line options, so we just need something like --determine-ip with appropriate explanation of what that means in terms of connecting to Syncplay.pl on the website and the --help descriptions.

Obviously there is then the possibility that someone won't be aware of the feature if they don't run --help and don't visit the website. One way around this would be to link people to http://syncplay.pl/guide/server/ in the welcome messages (e.g. "You are now running a Syncplay v{} server instance on port {}. Visit http://syncplay.pl/guide/server/ to learn more about running a server.")

abhsag24:
in addition to the welcome note, I also think  it would be a nice idea, to tell people to use --determine-ip,to know public ip, and --help to know more, at the place where it generally displays the network-ip.

Et0h:
Sounds reasonable, although if we're going that far it might be better just to list all of the command line arguments/options as there aren't that many of them.

abhsag24:
i'll get to displaying everything later right now it displays the earlier message

abhsag24:
also the constructor of  syncFactory in server.py is kind of getting messy, wouldn't it be better if we use *args,**kwargs instead of listing everything out?

Et0h:
The stubs for non-English messages should be the English message with a #TODO: Translate into Russian/German comment. I can merge and fix that if needed though, so long as people are otherwise happy with this PR.

abhsag24:
i tested the code through an NAT and found that the displayPublicIP() function, doesn't return the public ip, rather it only returns the ip through which, the current request was made, and that the only way to know the public ip is to query a web server. Keeping this in mind, we can 

1> Query a web service(since we already give an opt-in , we can tell the users that we will attempt to connect them to the web service), generally these services follow certain standards, and give the ip's for scripting purposes as well.

2> Query syncplay.pl and have the syncplay server return the ip .

3> We could leave the code as it is and just change the function name, as right now the message
reads "Your active internet ip is {}."  which is still the case, even though it wasn't exactly what we thought.

Uriziel:
Hey, I thought about your PR for a while and what if IP check actually was outside of main code but rather executed different routine upon `./syncplayServer.py --my-ip` displaying IP and exiting.

So it would be possible to check your IP whether your server is running or not + it would not change much of existing code.

Also I think querying syncplay.pl might be the easiest thing to do, we can set up action on webserver that'd return just that.

Et0h:
Have now created http://syncplay.pl/getipaddress/ which provides a JSON along the lines of:

{
your-ip: 1.2.3.4
}

abhsag24:
@Uriziel that would be cool! , though wouldnt it be unrelated to the syncplay software?

Uriziel:
It kinda is already, right?

abhsag24:
@Uriziel right now its a feature for the syncplayServer(part of the software) IMO

PR #90, opened 2015-12-26
Implement Chat: Clients able to send messages in a room

abhsag24:
refinements are still left such as, username is not displayed, the basic framework is done.

### DIFF
--- a/syncplay/client.py
+++ b/syncplay/client.py
@@ -454,7 +454,11 @@ class SyncplayClient(object):
         self.userlist.currentUser.room = roomName
         if resetAutoplay:
             self.resetAutoPlayState()
-
+           
+    def sendChat(self,message):
+        if self._protocol and self._protocol.logged:
+            self._protocol.sendChatMessage(message)
+        
     def sendRoom(self):
         room = self.userlist.currentUser.room
         if self._protocol and self._protocol.logged and room:
--- a/syncplay/constants.py
+++ b/syncplay/constants.py
@@ -60,6 +60,7 @@ FILENAME_STRIP_REGEX = u"[-~_\.\[\](): ]"
 CONTROL_PASSWORD_STRIP_REGEX = u"[^a-zA-Z0-9\-]"
 ROOM_NAME_STRIP_REGEX = u"^(\+)(?P<roomnamebase>.*)(:)(\w{12})$"
 COMMANDS_UNDO = ["u", "undo", "revert"]
+COMMANDS_CHAT = ["ch","chat"]
 COMMANDS_LIST = ["l", "list", "users"]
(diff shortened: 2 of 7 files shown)


abhsag24:
unicode text is working.

abhsag24:
The current problem in chat is, there is 
1.No UI to send Chat messages, right now.
2.if user types large messages, they are displayed on the screen, which disturbs the viewing experience

To tackle this problem we can 
a. Stop displaying messages in the player altogether 
or
b. Limit the length of message so that they are unobtrusive on the player (preferred)
or
c. Give the power to the client to stop displaying message on the screen in real time.

Uriziel:
You have to rebase your changes away from your other PR though.

abhsag24:
rebased the changes the away.

abhsag24:
another way we could stop large messages from displaying on the player is after a limit we could display only a certain part of the message, followed by "..." , rest of it could be read within the client or on the console

Uriziel:
How are you going to determine what's big message? (i.e. difference between fullHD and laptop screen)

abhsag24:
a full hd screen can have more characters , but displaying a part of the message would only serve as a kind of notification,  so displaying less characters on a bigger screen is not a big deal IMO.

remimarenco:
Hey @abhsag24 , would be glad to help you on this!

abhsag24:
@remimarenco would love to have help, as you can see i had already listed out the things that needed to be done further. What do you suggest?

remimarenco:
@abhsag24 We should take some time to discuss about your vision on your branch, and see how we can work together on this.
Personally, I just want to be able to chat with my loved one while we are watching the movie. Annoying to have to use a third party system or to speak.

I found the [letsgaze](https://letsgaze.com/) system pretty handy to handle the chat

perfectcolors:
I'm really looking forward to this functionality. I think having the conversation more closely tied with the on-screen action is essential for making it a more social experience.

Using Skype works and all, but I agree with @remimarenco that it's not always ideal to be on voice or to have to rely on another program to send text which requires keeping another window / screen active.

abhsag24:
@perfectcolors  @remimarenco i think i've added the basic chat functionality in this PR already, you can go ahead and use it, but you'll have to chat from within the terminal and there are some concerns that i had stated earlier, i don't know much about the GUI and don't have the time right now add that functionality :(  , but it should be pretty usable :)

Et0h:
Merged into https://github.com/Syncplay/syncplay/tree/chat

abhsag24:
awesome! so should i close this PR?

Et0h:
Closing this as PR has been accepted onto the chat branch.

PR #93, opened 2016-01-19
Shared playlists

Et0h:
* Add initial shared playlists feature
* Various minor tweaks and fixes


Et0h:
Is there anything that needs doing before this PR is accepted?

julthon:
Why is this still not accepted?

Et0h:
Outstanding issues that would ideally be resolved before release:
* (Major) At least one user has reported freezes/crashes when Syncplay changes files on mpv and MPC-HC. It might ultimately be the fault of the media players, but hopefully something can be done in Syncplay to avoid triggering the issue.
* (Moderate): Sometimes it skips two files rather than one.
* (Minor/Moderate) The 'highlight files not on PC in red' is not reliable enough, and sometimes gives false negatives/positives

Still to do (but doesn't have to be in initial release):
* Add more 'safe' streaming domains, and allow user to specify these
* Add looping functionality
* Allow media folders to be added during runtime, and prompt users for this when necessary
* Better handling for connecting to old servers without shared playlist functionality
* Ability to more easily save/load/copy shared playlists

Known existing/potential issues that are not under active development:
* Won't go to next file if the person who got to the end first is not using shared playlists - probably won't fix as it is not worth the hassle (if anyone cares, just get everyone to use shared playlists)
* Occasionally does not rewind after switching files - intermittent so hard to test, but might already be resolved
* Occasionally file selection may change when updating playlist - intermittent so hard to test, but might already be resolved

Et0h:
URI switching issue resolved:
* By default, if Shared Playlists is enabled people will be automatically switched to all URIs
* If people wish, they can set 'onlySwitchToTrustedURIs' to 'True' in '[client_settings]' within the Syncplay INI.
* Those users can then specify Trusted URIs via 'trustedURIs' in '[client_settings]'. E.g.:
`trustedURIs = [u'http://www.youtube.com/',u'https://www.youtube.com/']`

Et0h:
Hopefully resolved the 'highlight files not on PC in red' reliability issue with ea5813e2b5352ffb20802b10114d16214208aeb8 and 571f671ce2ffea519c6d1b7408873545297d5ac6

Et0h:
Hard to test as the issues are intermittent, but I have hopefully resolved the double advance issue with 	09907a0 and the 'not ready' EOF bug with f7b7855. Let me knows if these problems persist with the new version please.

Et0h:
All major issues should now hopefully be resolved. Thanks to everyone who has contributed to the shared playlist branch!

PR #94, opened 2016-01-25
Divide the international files

remimarenco:
Hi!

I would want to add the french language, but first I wanted to divide the international files into separate ones.
That way, each languages can be modified without taking the risk of damaging the others or the language controller (messages.py).

Let me know what you think about this ;).


Et0h:
Thanks for the suggestion, it will hopefully make translating a bit less cluttered. Looking forward to a French translation :)

Please note that we'll hopefully soon be adding messages relating to the new shared playlist feature.

Issue #57, opened 2015-03-25
Pause or reset to start on reaching end of file


wiiaboo:
Not really a bug, but an annoying feature on Syncplay keeps the same position even when a file ends or changes to the next on the playlist.
Happens with mpv and mpc-hc.
When the next file has a smaller duration it then instantly changes to the next until a file has higher duration and it finally stops.

Et0h:
I agree it is annoying. However, it is less annoying than if it didn't do it, because then if someone changes file during playback (e.g. because they had the wrong file open) it would reset everyone. This is also why it would mess up playlists, as the first person to get to the end of the file would prevent others from getting to the end of the file, and it is impossible for Syncplay to reliably determine whether you are near the end of the file (because not all APIs correctly report media duration).

If you shift+drag the file into Syncplay it is supposed to set position to 0, but it might not be reliable (programmers are free to try and improve the reliability of that functionality, it isn't something I use myself so it hasn't had full testing). Because of how Syncplay works it just can't be compatible with player-based playlists, but there are plans for a Syncplay-based playlist in the future - see https://github.com/Syncplay/syncplay/issues/40 for more details and to discuss these matters.

PR #97, opened 2016-03-22
added pause/set video to position 0 ON new file

nuthing:
# TODO: Add support for this setting in the GUI Config.

### DIFF
--- a/syncplay/__init__.py
+++ b/syncplay/__init__.py
@@ -1,4 +1,4 @@
 version = '1.3.4'
 milestone = 'Chami'
-release_number = '21'
-projectURL = 'http://syncplay.pl/'
+release_number = '22'
+projectURL = 'https://github.com/nuthing/syncplay'
--- a/syncplay/client.py
+++ b/syncplay/client.py
@@ -406,6 +406,9 @@ class SyncplayClient(object):
         filename, size = self.__executePrivacySettings(filename, size)
         self.userlist.currentUser.setFile(filename, duration, size, path)
         self.sendFile()
+        self.toggleReady()
+        self.updatePlayerStatus(True, 0)
+        self.setPosition(0)
 
     def __executePrivacySettings(self, filename, size):
         if self._config['filenamePrivacyMode'] == PRIVACY_SENDHASHED_MODE:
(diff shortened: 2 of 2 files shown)


syncplay/__init__.py
```diff
@@ -1,4 +1,4 @@
 version = '1.3.4'
 milestone = 'Chami'
-release_number = '21'
-projectURL = 'http://syncplay.pl/'
+release_number = '22'
+projectURL = 'https://github.com/nuthing/syncplay'
```
Uriziel: ??

nuthing: sry, was just added in for my project url since its not for public release.. i just wanted a place to post edits.
can you add support for authorURL and projectURL? i have no other options changed, just this url

Uriziel: 1. New feature is not new release - this value only changes once there's actual version-up
2. projectURL actually points only to Syncplay's website and is used as a message parameter somewhere.
3. what's the point of authorURL

nuthing: oh, i understand now i will fix in next push
authorurl was so i could test some code using git to update, ~ i guess i could put it in constants. i am still working on understanding the structure

Uriziel: Those values are not used in any areas regarding git or connections so you don't need to change them.

nuthing: i still think the other code i added was important. but i am still finding bugs like: when owner of room leaves and no one has permission to play, then the guest will keep jumping to position 0 forever until owner is back

Uriziel:
Also this behaviour is not necessarily wanted in all user cases so I'd be very reluctant to accept it as it is. 
Examples:
- User opened wrong episode, currently he can switch seamlessly
- User prepares next file while others are finishing old one
- User joins a room that already has file playing without resetting the state

Also why would you toggle someone's ready state after a file change? This is wrong.
Also "Shared Playlists" are solving most issues with switching files / rewinding too.

nuthing:
1 and 3 can be fixed with an option in the GUI to enable/disable.
The ready status was idea from the if user disconnect->pause, but in the case of file switch by accident but would need to be optional per situation to work.

Et0h:
Resetting the position to 0 when you hold shift while dragging in already exists in experimental form as per http://syncplay.pl/guide/client/ 'Main Window' -> 'Drag in file') and https://github.com/Syncplay/syncplay/issues/57 - I think if you want this feature it would be best to improve that functionality rather than add something that changes default Syncplay behaviour for dragging in a file (as this might result in inadvertent resets which is annoying).

Et0h:
In addition to this mostly already being a featute, it is also now made somewhat moot by the new shared playlist feature introduced in Syncplay 1.4.0.

PR #106, opened 2016-08-01
mpv bugfixes

wiiaboo:
scaletempo is also not useful with the small changes in playback rate from the slowing down in Syncplay. They're small enough to be more noticeable with it than without, like with interpolation+display-sync.
Besides, mpv has --audio-pitch-correction enabled by default which automatically adds scaletempo when needed.


wiiaboo:
--audio-pitch-correction was introduced in v0.9.0 and since then mpv has always dynamically added scaletempo as needed.

Et0h:
The change sounds like a good idea - I've assigned Dan to review the PR.

PR #111, opened 2016-08-22
Deprecate MD5 in favor of SHA512

Eriner:


Uriziel:
Hi,
while we're perfectly aware of weakness of md5 algorithm it's kinda non-issue for Syncplay server's password as it's meant to be shared anyway.

Merging your change has an issue of breaking backwards compatibility though.

If you consider this change to be important I'd like you to revise your change so server accepts either hash.

Eriner:
> while we're perfectly aware of weakness of md5 algorithm it's kinda non-issue for Syncplay server's password as it's meant to be shared anyway.

It would be ideal if the password was not reversible from an intercepted connection, even if the password is known my many individuals. That shared password may be used other places by the same group of individuals, which may allow an eavesdropper to access other services.

> Merging your change has an issue of breaking backwards compatibility though.
>
> If you consider this to be important I'd like you to revise your change so server accepts either hash.

Can do. I will update this PR shortly. Perhaps this is a change that can be implemented in both the client and server for 1.4 before it goes stable (seeing as it has breaking changes from 1.3). Unless I am incorrect about the incompatibility between the versions, would it not be ideal to enforce SHA512-only authentication in 1.4+ ?

Uriziel:
> Unless I am incorrect about the incompatibility between the versions, would it not be ideal to enforce SHA512-only authentication in 1.4+ ?
It's fine if it stays as new clients will be sending SHA512 version anyway and MD5 won't be stored / exchanged anywhere except if older client requires it.

AFAIK Some people don't care about updating Syncplay whatsoever.

Eriner:
> It's fine if it stays as new clients will be sending SHA512 version anyway and MD5 won't be stored / exchanged anywhere except if older client requires it.

Right, but isn't there an incompatibility between 1.3 clients and the 1.4 server? Am I missing something? If there is an incompatibility and this change will only land for 1.4, it seems to make sense to only allow SHA512 authentication. 1.3 clients (sending MD5) are incompatible with the 1.4 server anyway (unless I am wrong about the incompatibility)

Uriziel:
They're compatible.

Eriner:
Ah, that clears the confusion. I will implement the check for both hashes and update this PR.

Et0h:
1) All public version of Syncplay 1.* client and server are backwards and forwards compatible. Syncplay is premised on both backward and forwards compatibility - clients need to be compatible with older versions of the server, not just servers needing to be compatible with older clients. Our intention is that backwards compatibility will only be broken when we move to Syncplay 2.

2) People should never re-use a Syncplay password. I've now made this explicit in the server guide.

3) Simply moving from MD5 to SHA512 would mean it would still be susceptible to a repetition attack and would still have no salt (making it easier to crack).

4) If you wanted better security/privacy then it would be worth implementing end-to-end encryption of all of the communications. This is something I hope to see in Syncplay 2.

Eriner:
> 1) All public version of Syncplay 1.* client and server are backwards and forwards compatible. Syncplay is premised on both backward and forwards compatibility - clients need to be compatible with older versions of the server, not just servers needing to be compatible with older clients. Our intention is that backwards compatibility will only be broken when we move to Syncplay 2.

I have updated the PR to make it backwards compatible with older clients.

> 2) People should never re-use a Syncplay password. I've now made this explicit in the server guide.

Agreed.

> 3) Simply moving from MD5 to SHA512 would mean it would still be susceptible to a repetition attack and would still have no salt (making it easier to crack).

Yes, this isn't the best move, but it was the easiest way to add some additional security with a minimal change to the codebase and without introducing any additional module dependencies (such as `bcrypt`).

4) If you wanted better security/privacy then it would be worth implementing end-to-end encryption of all of the communications. This is something I hope to see in Syncplay 2.

This is the ideal solution, but as you said, this would for sure break compatibility.

This patch is not meant as a long-term fix to the lack of full encryption, only a short-term mitigation

Uriziel:
> 3) Simply moving from MD5 to SHA512 would mean it would still be susceptible to a repetition attack and would still have no salt (making it easier to crack). 
No, it doesn't as there's no repetition there (you send either SHA512 or MD5)

Et0h:
> I have updated the PR to make it backwards compatible with older clients.

It isn't just about allowing the server to be compatible with older clients, but also about allowing clients to be compatible with older servers. As far as I can see this change would prevent newer clients from connecting to older servers that are password protected.

Eriner:
> It isn't just about allowing the server to be compatible with older clients, but also about allowing clients to be compatible with older servers. As far as I can see this change would prevent newer clients from connecting to older servers that are password protected.

Ah, of course. I forgot to add this one. I will send another commit shortly.

Et0h:
> No, it doesn't as there's no repetition there (you send either SHA512 or MD5)

By repetition I'm talking about a replay attack of the SHA512, not it being repeated in MD5.

Uriziel:
> As far as I can see this change would prevent newer clients from connecting to older servers that are password protected.
Dunno if it's really necessary though.

Eriner:
Actually, no, sending the MD5 defeats the purpose. I don't know if there is a non-compatibility breaking change that provides any real security here if we have to send both to maintain compatibility. Sending the SHA512, checking if it's valid, then sending the MD5 if it fails still "leaks" the password.

Uriziel:
I agree with @Eriner here. PR is good enough for me.

Et0h:
> PR is good for me.

I'm not happy losing compatibility with older servers, especially for something that would not be a great leap forwards. A compromise would be that it only repeats the password if it is connected to an old server. I think to do this the server would have to send the server version along with the 'wrong password' error message and the absence of the server version would indicate to the newer client that it was an MD5-only server.

> You can replay MD5 too. Unless some kind digest-three-way-handshake auth is done you're not going to mitigate that.

I didn't say it was a new issue, but you are correct that resolving the issue would require a more complicated handshake protocol.

Eriner:
> the server would have to send the server version

If this is the case, why not just kill this PR and implement a (temporary) opportunistic "upgrade" to end-to-end crypto?

The client sends the following to the server upon first connect:
```
{"Hello": {"username": "Eriner", "password": "acbd18db4cc2f85cedef654fccc4a4d8", "room": {"name": "mumble"}, "version": "1.2.255", "realversion": "1.3.4"}}
```

Assuming I understand your proposal correctly, that being flow control for crypto (SHA512 vs MD5) previously, I think it best to use the flow control to upgrade the connection to an encrypted one. After a successful connection, the client could automatically toggle some boolean associated with the configuration to "remember" the upgrade. Should the server ever be downgraded and the full-encryption lost, the client can reset or uncheck that box.

This seems like the best solution. I believe going this route would kill two birds with one stone: stronger crypto now for clients/servers that support it, and backwards compatibility with older clients/servers.

What are your thoughts @Et0h, @Uriziel?

Et0h:
I'm perfectly happy for someone experienced with both encryption and Python to add end-to-end encryption to Syncplay so long as they can do it without unacceptably impacting on performance or resource use for the server or client, but E2EE it isn't something I have any experience in so I am in no hurry to try to do it myself.

diogocp:
> Yes, this isn't the best move, but it was the easiest way to add some additional security with a minimal change to the codebase and without introducing any additional module dependencies (such as `bcrypt`).

`hashlib` supports PBKDF2. See https://docs.python.org/2/library/hashlib.html#key-derivation.

FSMaxB:
I know this issue is kind of old, but I think there is no benefit to this change at all, unless you switch to something like pbkdf2 or bcrypt, because SHA512 is about as weak as MD5 is against the usual kind of attack on passwords: Brute Force Attacks. You could argue that there are more rainbow tables for MD5 than SHA512, but in the end, switching to SHA512 fixes nothing if the underlying mechanism is flawed.

Eriner:
> but in the end, switching to SHA512 fixes nothing if the underlying mechanism is flawed.

Correct, the real fix here is to create a backwards-compatible secure connection upgrade negotiation, wherein the clients can enforce a secure connection (or remember the connection a la HSTS), ideally using a password for authz and then switch to cert-based communications. I'm closing this PR, as stated previously, this was only a minor minor patch to at least move away from MD5 without introducing any additional dependencies; this was never meant as a real security fix.

PR #117, opened 2016-10-13
Try to setup GUI, iff noGui is False

Hummer12007:
This prevents the import error message from being displayed when GUI is
disabled on systems without PySide

Additionally, remove some redundant conditionals

### DIFF
--- a/syncplay/ui/ConfigurationGetter.py
+++ b/syncplay/ui/ConfigurationGetter.py
@@ -7,13 +7,6 @@ from syncplay import constants, utils, version, milestone
 from syncplay.messages import getMessage, setLanguage, isValidLanguage
 from syncplay.players.playerFactory import PlayerFactory
 import codecs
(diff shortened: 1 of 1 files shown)


Et0h:
Will need to test this before it is merged, but thanks for your contribution. It is definitely desirable to allow those without PySide to use the command line interface version of Syncplay.

Et0h:
Only quickly tried to test, but I couldn't get it to work. It seems that when you run it and then have GuiConfig load by default and then run Syncplay from GuiConfig it adds an additional :8996 to the end of the host/ports which causes Syncplay to fail to connect.

Not sure, but I think it is related to the fact that the new code causes _validateArguments to be called twice BEFORE the GuiConfig dialogue is closed and not at all afterwards.

Have you tested it with forceGuiPrompt being true in your config file and PySide installed?

PR #118, opened 2016-11-02
Add noGui configuration option to config reader

Hummer12007:


Et0h:
I have no problem with people being able to set noGUI via their INI if they wish, but I'm concerned that it might result in that value being set to "true" in the event of PySide not being installed and then subsequently stay disabled once PySide is actually installed. So loading from INI is fine, but saving to INI may not be the correct behaviour. Will need to look into how the code operates to see if this is an issue.

Hummer12007:
Solved it downstream (by patching the default value, when GUI is disabled). Closing.

PR #120, opened 2016-11-23
Support mpv without length property

wiiaboo:
The 'length' property was deprecated since then and will be gone
with 0.23.0 and is already gone in git master, so Syncplay is
broken with it.


Et0h:
If this change to Syncplay breaks backwards compatability then it raises
the question of whether it can be modified to maintain backwards
compatability, e.g. by having different behaviours for different versions.

wiiaboo:
If they're using a more-than-a-year-old mpv then they can use a more-than-a-year-old Syncplay too.

wiiaboo:
This is the commit that deprecated it, from May 2015: https://github.com/mpv-player/mpv/commit/d26803cba00f9ecfe7c1c0914f133e775638183b

Et0h:
Whilst I would rather everyone ran the latest versions of software I also
aim to maximise backwards compatability. As such, I would be reluctant to
opt for a change that broke backwards compatability when an alternative
which maintained backwards compatibility required little additional effort
and had no adverse side effects.

Of course, forwards compatibility is also very important so thanks for
proposing this change. I shall look into resolving the pull request when I
return from my holiday in a few days if neither Uriziel nor Dan has
resolved it by then.

wiiaboo:
Hey, turns out we can use both. And livestreams/files with unknown durations work now.

Et0h:
Excellent ^_^

wiiaboo:
Merged locally?

Et0h:
>_<

Issue #146, opened 2017-09-15
[OS X] Can't add files to shared playlist.


Duamer:
This problem occurred in two different systems I have. A clean slate system that only had Syncplay installed and my main, lived-in system.

I can'd add files to the shared playlist. Neither drag-and-drop, nor right-click, "Add file(s) to bottom of playlist" function. I momentarily see a files browser window pop up when I right-click.
I can add URLs to the playlist and they work as they should. I've setup the media directories and had a different user make the playlist and that works as well.

alby128:
Confirmed. Looking for a workaround. Thanks.

alby128:
If you have time, please test the workaround that I have just pushed to my branch: https://github.com/alby128/syncplay . Right now, I tested it on 10.12.6 and it works for me.

Duamer:
How do I install pyobjc with homebrew? I used the homebrew method of your directions.

Incase that's not what's causing my error, here's what I got:

users-iMac:syncplay user$ python2 syncplayClient.py
2017-09-15 19:22:39.429 Python[628:5526] modalSession has been exited prematurely - check for a reentrant call to endModalSession:
Traceback (most recent call last):
  File "syncplayClient.py", line 19, in <module>
    SyncplayClientManager().run()
  File "/Users/user/syncplay/syncplay/clientManager.py", line 9, in run
    interface = ui.getUi(graphical=not config["noGui"])
  File "/Users/user/syncplay/syncplay/ui/__init__.py", line 9, in getUi
    ui = GraphicalUI()
NameError: global name 'GraphicalUI' is not defined
Segmentation fault: 11

alby128:
To install pyobjc on your homebrew, just run `pip2 install pyobjc`. However, I have to admit that your error message puzzles me a little... Where and how did you get it? When you tried to add something to a shared playlist?

Duamer:
Looks like that fixed it. I'll be able to test it by Monday to see if it works with other users in the room.

I got that error when I clicked "Store configuration and run Syncplay."

alby128:
@Duamer Did you test this fix with other users?

Duamer:
Testing now and it looks good. I was able to setup the room and the order user joined. Works like it does when I was using the windows client.

Duamer:
Testing now and it looks good. I was able to setup the room and the order
user joined. Works like it does when I was using the windows client.

alby128:
[...]

PR #148, opened 2017-09-15
Fixes issue #146

alby128:
On macOS, fixes the "Add to the bottom of shared playlist" QFIleDialog with the `QtGui.QFileDialog.DontUseNativeDialog` workaround used before. Also fixes drag and drop of media files on both the "List of who is playing what" and "Shared playlist" boxes.

Drawbacks: requires PyObjC. If merged, the installation instructions should be updated to include `pip install pyobjc`.

### DIFF
--- a/syncplay/ui/gui.py
+++ b/syncplay/ui/gui.py
@@ -11,6 +11,8 @@ import os
 from syncplay.utils import formatTime, sameFilename, sameFilesize, sameFileduration, RoomPasswordProvider, formatSize, isURL
 from functools import wraps
 from twisted.internet import task
+if sys.platform.startswith('darwin'):
+    from Foundation import NSURL
 lastCheckedForUpdates = None
 
 class UserlistItemDelegate(QtGui.QStyledItemDelegate):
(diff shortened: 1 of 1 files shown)


Et0h:
http://www.birving.com/blog/2015/09/29/drag-and-drop-files-into-gui-using/ indicates one should install pyobjc-core before installing which may relate to the issue at https://stackoverflow.com/questions/36813269/issue-with-installing-pyobjc-for-python but I'm not sure. As far as I can tell, if we're using PySide 1 we need to use NSURL to support file dragging and that means we need PyObjC. Are there any major drawbacks to requiring it other than requirement bloat?

alby128:
That birving.com page is what I used as reference to write this workaround.  You are right, we need NSURL and I did not find an alternative that does not rely on PyObjC. I do not see any particular problem in requiring it in the instructions, my concern was about the size of PyObjC. Including it in a .app bundle should increase the size of the packaged file.

alby128:
Just an FYI, the current master of Syncplay, when packaged in an .app, is about 39.9 MB. The inclusion of the `Foundation` module from `pyobjc` increased it to about 41.4 MB. So I guess that was too much ado about nothing.

Issue #89, opened 2015-12-23
OS X: package into .app


blaenk:
See [this](http://stackoverflow.com/a/34161466/101090) SO answer by someone who surveyed a couple of options. This will help solve #87

alby128:
Fixed by PR #150 .

Issue #87, opened 2015-12-23
OS X: change "python2" menu bar title to "Syncplay"


blaenk:
Issue #86 fixes the menu bar on OS X, which is awesome, but it shows "python2" instead of "Syncplay" as the application title.

It looks like in order to change the application name from "python2" to something else, we need to edit some `Info.plist` file in the generated .app (if we're even generating one at the moment). [Source](http://stackoverflow.com/questions/7827430/setting-mac-osx-application-menu-menu-bar-item-to-other-than-python-in-my-pyth).

Alternatively, this [answer](http://stackoverflow.com/a/34161466/101090) in particular seems to go over packaging methods for python apps on OS X.

nilsding:
This issue will be fixed once we manage to package Syncplay as an .app.  I'll take a look at generating one using _PyInstaller_.

blaenk:
Yep. Check out the last link, the guy surveys a couple of options including PyInstaller.

alby128:
Fixed by PR #150.

PR #150, opened 2017-10-14
Packaging of Syncplay in a single executable .app file for macOS

alby128:
Gives a way for users on macOS to run Syncplay without evening opening their terminal once. Creation of the .app and packaging of it in a distributable .dmg are handled by Travis CI. Appveyor support is also experimentally added for Windows. Binaries deployment is done through Bintray.

Also, solves issues #87 and #89. In addition, it adds an About dialog on all operative systems that include information about Syncplay, its license and the license of its dependencies.


Et0h:
Does  "git clone -b macOS_app https://github.com/alby128/syncplay.git syncplay_app" needed to be amended to use official Syncplay repo?

Et0h:
And does the bintray name still want to be "1.5.0_test"?

alby128:
Fixed and fixed. Changed bintray version to 1.5.0, even though bintray will not be used (probably) for officially distributing the .dmg file. If we decide to do so, we should probably create a bintray account for Syncplay.

PR #165, opened 2017-12-25
buildPy2exe: Find installed NSIS through registry

wiiaboo:
Require NSIS >=3 for unicode support


alby128:
In OUT_DIR, there was an underscore between syncplay and v in the unchanged version. I used that for appveyor tests. Could you please put that _ back (if that is not a problem for anything else)? Thanks and merry Christmas :)

wiiaboo:
I guess that's why there was a merge conflict. Done.

PR #176, opened 2018-02-08
Supporting for Italian translation.

Nico769:
The entire application and its installer have been translated into Italian.

A hard coded label in **_ui/GuiConfiguration.py_** has been replaced by a new label `chatoutputheader-label`. Thus, `self.chatOutputGroup = QtWidgets.QGroupBox(u"Chat message output")` is now `self.chatOutputGroup = QtWidgets.QGroupBox(getMessage("chatoutputheader-label"))`.

For clarity's sake, `chatoutputheader-label` has also been added to the others **messages_xx.py**

### DIFF
--- a/buildPy2exe.py
+++ b/buildPy2exe.py
@@ -57,6 +57,7 @@ NSIS_SCRIPT_TEMPLATE = r"""
   LoadLanguageFile "$${NSISDIR}\Contrib\Language files\Polish.nlf"
   LoadLanguageFile "$${NSISDIR}\Contrib\Language files\Russian.nlf"
   LoadLanguageFile "$${NSISDIR}\Contrib\Language files\German.nlf"
+  LoadLanguageFile "$${NSISDIR}\Contrib\Language files\Italian.nlf"
 
   Unicode true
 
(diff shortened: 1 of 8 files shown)


alby128:
I am sorry, I did not notice that Travis CI builds were broken for syncplay/master. I just fixed this with the `394fd6e` commit. @Nico769 could you please update your fork (hence this PR) by adding this commit? Thanks. I will then review all the translated labels. 

For @Et0h: for some reasons I did not get an email notification about the broken build on Travis, I am looking into this right now. Also, Travis is still deploying to version v1.5.1 on Bintray. I suggest to just get rid of versions on Bintray and stick with something like "stable" and "development", so that we do not have to update the bintray.json file everytime.

Et0h:
@alby128 Makes sense. Is that something you can change or do I need to do it?

Nico769:
@alby128 No worries. It should be ok now.

alby128:
@Et0h I will just edit the "bintray.json" file to use "development" as version name. Then, we will just manually move the correct files in the "stable" version (that I will create manually) once they are ready. I think keeping one or two stable versions on bintray should be more than enough. Also, right now appveyor builds are failing because there is not enough space to upload artifacts. We have to delete some artifacts from the appveyor Syncplay's account, but unfortunately, I cannot do that from my own account.

@Nico769 I will review the Italian translation in the next days. Thanks.

Et0h:
@alby128 What would I need to do to delete some artefacts? I tried deleting some of the old builds but that didn't make a difference, but perhaps I just need to delete more?

albertosottile:
@Nico769 I am reviewing the translation and I think it will be easier if you let me commit edits directly in your fork. Could you please enable the write access as described here: https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/ ? Thanks.

albertosottile:
@Et0h I also need some help about this:
`"mpv-key-tab-hint": u"[TAB] to toggle access to alphabet row key shortcuts.", # TODO needs to clarify this` (line 261)
this:
`"fastforwardondesync-tooltip" : "Jump forward when out of sync with room operator (or your pretend position if 'Never slow down or rewind others' enabled).",` (line 371)
and this:
`"not-known-server-error" : "You must be known to server before sending this command",` (line 446).

Also, I would like to know what the "Hello" arguments are before attempting to translate the relative strings.

albertosottile:
@Et0h one more thing:
Maybe we should avoid using `syncplay-mediasearchdirectories-title` also for the text above the input field and create a new label for it. Currently, this label is too long for the title bar but is fine for the dialog.

Nico769:
@albertosottile The option is already enabled.

Et0h:
@albertosottile Thanks for separating the messages. It was only for convenience that they were one message in the first place, but I can certainly see why it is worth splitting them to allow for more effective translation.

albertosottile:
The latest two commits conclude my review of the translation, nice work @Nico769. I will merge this PR as soon as I able to re-test everything in the app.

Nico769:
@albertosottile Thanks, you too!

PR #179, opened 2018-03-09
Fixing some unicode and Italian translation issues

albertosottile:
I am experimenting with files with non-ASCII characters in their name or path, and I have found some issues in the current version. In the following, there is a detailed list of what I found so far and how I fixed it. I am using this PR to keep track of these bugs and changes, since I am currently in the process of finding and debugging some of them.

Unicode support
- [x] file properties (name, size, duration) are not loaded if filepath contains non-ASCII characters.
- [x] (macOS) drag-and-drop of files with non-ASCII characters in their name does not open the file.
- [x] (macOS) drag-and-drop of files with non-ASCII characters on the shared playlist does not load the file.
- [ ] (VLC) "file differences: name" is initially triggered when a file with non-ASCII characters in its name is opened from VLC, then the player drops this difference, but the OSD does not update. (These may be more bugs, actually).

Italian translation
- [x] (VLC) Readiness status is not shown in the OSD if there are any file differences.

### DIFF
--- a/syncplay/client.py
+++ b/syncplay/client.py
@@ -1929,6 +1929,7 @@ class FileSwitchManager(object):
             return False
 
     def notifyUserIfFileNotInMediaDirectory(self, filenameToFind, path):
+        path = path.decode('utf-8')
         directoryToFind = os.path.dirname(path)
         if directoryToFind in self.mediaDirectoriesNotFound:
             return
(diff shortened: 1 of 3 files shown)


Et0h:
I'm hoping to release 1.5.3 soon to fix an issue with HTTPS on the Windows builds. Is this okay to merge or does it need more tweaks to resolve the outstanding issues?

Issue #184, opened 2018-05-05
MacOS - Change Language - crash due to untrusted SSL certificate


Shakal1710:
Hello,
I changed the language from English to German.
After a restart of syncplay the following error occurs:
<img width="420" alt="error" src="https://user-images.githubusercontent.com/39008459/39664509-75d8d89c-5084-11e8-8eec-8a1c48a978bb.png">

can I change the language back manually in a file or something else?

OS: MacOS High Sierra 10.13.4
syncplay: last (1.5.3a)

edit:
deleting and reinstall of syncplay do not change anything

Et0h:
I don't know how it works on macOS, but on other platforms you can resolve this by either (a) running the client with the --language en commandline switch as per https://syncplay.pl/guide/client/ or (b) delete the .syncplay (or on Windows Syncplay.ini) file. Not sure why German didn't work in the first place though - was there any more debug information?

Shakal1710:
ahh nice, thanks...
i have found an hidden file in my User-Folder :-)
/Users/****/.syncplay

i have changed the language back to en. Now it works :-)


hm i have found this in the apple-console: May  6 18:26:17 ****-iMac com.apple.xpc.launchd[1] (pl.syncplay.Syncplay.17036[2757]): Service exited with abnormal code: 255

albertosottile:
Thanks, I found and fixed the bug, but many questions arise:
- Why was the client unable to retrieve the list of public servers? Are you using a firewall to block Syncplay?
- Why did the Syncplay app crash after that traceback? The client is supposed to attempt sending the message and continue, which is what happened during my tests. The client never crashed. I suppose there is at least another bug that is triggering a "Syncplay Error" in the last line, but I am afraid that until we push a release with the first fix, you will not see a more detailed error message about the crash. @Et0h do you have any suggestions?

Note on how I fixed the bug: given that all the message strings are already Unicode, I dropped the cast and used e.args[0] as the argument for the showErrorMessage() function.

albertosottile:
Actually, I found a way to move this further. @Shakal1710 could you please download and install Syncplay from my bintray account? [This is the .dmg](https://bintray.com/alby128/Syncplay/download_file?file_path=Syncplay_1.5.4.dmg) that contains the fix mentioned above (plus a lot of other new stuff that hopefully will not interfere with this bug). After you installed the updated app, run it from Terminal again and see if you can still reproduce the crash. If so, post here the (hopefully updated) error log. Thanks again.

[...]

Issue #152, opened 2017-10-18
Ideas for migrating to Qt5


alby128:
This issue serves as a discussion and as a reference about migrating the Syncplay codebase to Qt5. 
The release of version 1.5.0 should happen soon, finally supporting macOS with a packed .app and no need of manual installation. Once that is settled, we should have some time to consider this migration.

In my opinion, there are three alternatives at this moment:
- defer this migration and stick with PySide1.x
- directly porting all the UI code to PySide2
- port the UI code to [Qt.py](https://github.com/mottosso/Qt.py). 

All these strategies have pro and cons, I will try to list them hereafter, being as objective as possible. In the following, I did not include PyQt5 because its open-source license (GPL) is not compatible with the current license of Syncplay.

### Defer the migration
- **Pro**: postponing is always easy, no changes required.
- **Cons**: Qt4 is not supported on any operative system anymore. We already have several patches specifically designed to correct quirks of Qt4 on macOS 10.12- (see PR #145 and #148). Plus, I have not tested Syncplay on the latest version of macOS yet. Any major OS update could break Syncplay entirely. This is not hypothetical, as it happened a year ago with macOS 10.12, see issue #116.

### Port the UI code to PySide2
- **Pro**: porting everything to PySide2 is not that difficult, due to its syntax. In fact, a working port is already available on [alby128/pyside2](https://github.com/alby128/syncplay/tree/pyside2), though it has not been updated to the latest commits yet. Doing this will keep the complexity of the codebase of Syncplay almost unchanged.
- **Cons**: while PySide2 works with the current codebase of Syncplay, it is not ready for production deployment. PySide2 is still under active development and its team has not even released a tagged version so far. in this situation, finding binaries of PySide2 is quite hard, and sometimes they are not even consistent between each other, because they have been compiled with quite different sources.

### Port the UI code to Qt.py
Note: [Qt.py](https://github.com/mottosso/Qt.py) is a shim that allows to code the UI using a common syntax and supports all the major Qt bindings (PySide1, PySide2, PyQt4, and PyQt5).
[...]

Issue #149, opened 2017-09-19
[OS X][UI Request] Add a /Volumes/ shortcut.


Duamer:
Small request. Add a way to easily get to the /Volumes/ directory in the "Set media directories" interface. My workaround was to manually type the path, but many users wouldn't think to do that.

I would suggest adding a shortcut in the "Find Directory" window and an exception so that same window can see the directory in the root of the drive.

alby128:
From a strictly technical point of view, this is possible, although it requires a fairly large amount of code rewriting. The self-initializing function getExistingDirectory needs to be replaced with a custom QFileDialog, encapsulated in a new method, and manually designed to behave in the same way, so to select only directories. Then, the /Volumes/ path could be added in it by using setSidebarUrls.

But, we should not forget that this issue was actually created by the use of non-native QFileDialogs instead of native dialogs, and that we had to introduce it as a workaround for some obscure qt4 bug on macOS. In macOS native file dialogs, this issue does not exist because the Finder shows all the mounted volumes in its sidebar.

Fixing this issue will require writing a substantial amount of macOS specific code that
- could stop working in the next major macOS release in a few days
- violates Apple's Human Interface Guidelines as it exposes the /Volumes/ folder to the GUI
- will become obsolete the day Syncplay migrates to qt5.

So, at this moment, I think this issue should not be fixed. I am of course open to suggestions.

Et0h:
@alby128 @Duamer Now that we have Qt.py support is this fully resolved?

alby128:
It will be resolved if and only if he uses PySide2 as binding.

Et0h:
@alby128 Okay, so I guess that before closing this issue we need a short sentence explaining the problem and how it can be resolved for the Troubleshooting guide at http://syncplay.pl/guide/trouble/

alby128:
@Et0h I tend to agree, although we are not providing anymore a guide to manually install Syncplay on macOS, so this possibility will be available only for experienced users. Maybe we should just leave this issue open until we switch to PySide2 in packaged .exe and .dmg .

Et0h:
@alby128 Happy to keep this open until we have it packaged in .dmg. Not sure why there would be a need for it to be packaged in the Windows .exe, though.

alby128:
@Et0h I just thought you wanted to keep all the platforms on the same binding, but that is not really necessary, you are right.

PR #185, opened 2018-05-07
Transition to PySide2 and Qt 5.x

albertosottile:
This PR contains the migration of Syncplay to PySide2, following the discussion of issue #152.

While it is true that PySide2 (side note: the official name is now Qt for Python) has not been officially released yet, the binding is stable enough for beta testing its behavior with Syncplay. Furthermore, in view of the upcoming dismissal of Python 2, we need to accelerate the transition to Qt 5. Ideally, once this PR is merged, we will release a beta 1.5.4 version to allow extensive tests of Syncplay on Qt 5 on all the platforms.  

Despite the large number of commits indicated by GitHub, only the packaging files are changed in this PR, while the Syncplay code is actually the same as in current master, thanks to Qt.py (with the exception of a few tabs in ui/ConfigurationGetter.py).
 
The code in this PR updates the packaging CI systems as detailed hereafter:
- Travis: embeds a [custom wheel](https://github.com/albertosottile/pyside2-macos-linked/tree/qt59) of PySide2 5.9 and Qt 5.9.5 in packaged .dmg for macOS.
- AppVeyor: embeds conda binaries of PySide2 5.6 and Qt 5.6.2 in packaged setup .exe and portable .zip for Windows.
- AppVeyor: embeds a blank syncplay.ini and an instruction file for LUA scripts in the portable release for Windows.

Incidentally, this PR solves issue #149.

Known limitations:
- this PR drops the support of macOS 10.10 due to the lack of a bottled version of Qt 5.9 for that operative system. This issue is mitigated by our legacy version that has already been released and supports 10.6.8 and following. 
- on Windows, we are still using Qt 5.6 due to the lack of a conda binary of PySide2 for Qt 5.9. It would be better for us to embed Qt 5.9 on both platforms since it will be supported at least for the next two years. This issue will most likely resolve itself when PySide2 approaches an official release. Otherwise, in the future, we could start experimenting with official PySide2 wheels on Windows.
- on macOS, custom wheels are needed because py2app does not work with official PySide2 wheels, which integrate their own version of Qt. This issue is mitigated by a [custom script](https://github.com/albertosottile/pyside2-macos-linked/tree/qt59) that currently packages the wheels used in Syncplay.


albertosottile:
I just added three commits to fix issue #184, add a Qt 5 license notice, and fix a typo in all the messages.

albertosottile:
I just added two commits to fix issue #184 (for real this time) and add a license notice for [Requests](http://docs.python-requests.org/en/latest/index.html) (Apache 2). @Et0h did you make any tests on Windows for this PR or are you aware of test results from other people? When do you think we should merge this?

albertosottile:
Side note: I tried to embed the PySide2 official wheels on Windows, but they are not releasing them for x86 and, as far as I know, there are no wheels for py2exe_py2 for win_amd64. So, for the moment, we are stuck with conda packages. We could attempt this replacement after migrating the codebase to Python 3 and PyInstaller.

albertosottile:
Updated after merging PR #188.

albertosottile:
Qt for Python was [released today](https://blog.qt.io/blog/2018/06/13/qt-python-5-11-released/), as a technical preview (v5.11.0 TP), with some caveats:
- Only Qt 5.11 and onwards will be supported
- Official wheels can be downloaded from [here](http://download.qt.io/official_releases/QtForPython/pyside2/) (they are waiting to include them in PyPi)
- As you can see, there are no wheels for cp27-win32 and apparently they will not release them.

In summary, this is the status of this PR for the three platforms:
- **Windows**: we are still using a (quite old) PySide2 5.6 package from anaconda. There is no way to use PySide2 official wheels. We can either build our own wheels or wait for the anaconda guys to update their packages.

- **macOS**: the latest commits embedded PySide2 v5.11.0 TP in the binaries, so everything should be ready to release a beta (note for @Et0h: we are using custom-built PySide2 wheels with the segfault commit reverted, as suggested by the PySide2 team).

- **Linux**: the codebase is ready for testing and now PySide2 is installable directly from official wheels by running `pip install --index-url=http://download.qt.io/official_releases/QtForPython/ pyside2 --trusted-host download.qt.io` so, I encourage everyone to test it. I suggest to release the beta of Syncplay and update the Linux  installation instructions to include PySide2 official wheels, instead of PySide 1.x.

albertosottile:
For the moment, we decided to stick with PySide 1.x on Windows, and migrate to PySide2 v5.11.0 on the other two platforms. The PR has just been updated accordingly.

Et0h:
@albertosottile Am I correct that the primary change of this PR is to move the macOS default build from PySide1 to a Syncplay-compatible version of PySide2 and that after this PR Linux users should continue to be able to use either PySide1 or PySide2?

albertosottile:
Yes, perfect recap.

PR #188, opened 2018-06-05
Change config path logic to support XDG

sometoby:
Someone in IRC helpfully remarked that:
>~/.syncplay belongs in XDG_CONFIG_HOME

Here's a PR that makes the Syncplay client a little more XDG-conscious. I refactored the related code to aid my understanding of it, and to simplify the subsequent XDG changes.

https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html


albertosottile:
@sometoby Can you close and reopen this? I want to see if, after the changes that I applied to the Travis script, it passes the check. Thanks.

sometoby:
Your commit seems to have already triggered the CI, I think we're good.

Et0h:
Thanks for your work on this. The config file does indeed belong in XDG_CONFIG_HOME where appropriate.

I am not familiar with XDG_CONFIG_HOME so I cannot rule out the possibility that on some non-NT systems there are circumstances where the XDG_CONFIG_HOME environmental value is set. It might be safer for the _getXdgConfigHome code to confirm that XDG_CONFIG_HOME exists (and is not empty/blank/null) before using XDG_CONFIG_HOME to create a directory and return a path.

I tested it on Windows 10 and it was fine both when there was and wasn't an existing .ini file. Not tested Linux, macOS or *BSD.

@albertosottile @daniel-123 @nilsding Can you confirm that the change works okay on macOS, Linux and *BSD respectively when finding an existing config file in the old style and when creating a new config file?

nilsding:
@Et0h the change works fine on FreeBSD; it uses `~/.syncplay` if it exists and `$XDG_CONFIG_HOME/syncplay.ini` otherwise.

albertosottile:
@Et0h I can confirm that everything works as designed on macOS. It uses ~/.syncplay if available or creates and uses ~/.config/syncplay.ini if there is no previous configuration file in the home folder. AFAIK, $XDG_CONFIG_HOME is never set on macOS but, if the user sets it manually (not a so trivial task in this OS), then Syncplay behaves accordingly.

Issue #169, opened 2018-01-08
Python 2 End of Life 2020


FSMaxB:
Python 2 will [stop receiving support in 2020](https://www.python.org/dev/peps/pep-0373/). Are there any plans already to port syncplay to Python 3?

alby128:
Just dropping my 2 cents here, nothing official. We are currently in the middle of another transition, changing the GUI binding to support (eventually) Qt 5. So, in my opinion, it would be better to wait for the official release of PySide2 before thinking to migrate to Python 3. Hopefully, this should happen long before 2020.

That being said, I played a little with 2to3 and syncplay today, and the most annoying compatibility issues that I found so far were caused by the str/Unicode/byte type changes introduced in Python 3. These issues will force to adjust all the player protocols introducing a lot of decode() and encode() in sent or received strings, according to the types that Twisted for Python 3 is now expecting.

This is what I found so far, but maybe someone else is aware of further, much more worrisome, potential problems.

xNinjaKittyx:
I'm playing around with syncplay on python3.6 and so far I've been able to easily setup the server and have users connect to it. Chat seems to work, and now I'm revising mpc.py since the connections are being disconnected when the server updates the client what timestamp it's supposed to be.

Also, I just want to point out something funny I found while looking through the code.
![image](https://user-images.githubusercontent.com/1697670/34808797-088908ae-f646-11e7-847b-56514ad7e7c2.png)

Et0h:
Good question, @FSMaxB. Python 3 support is not something that I have looked into for many years and no recent decisions have been made as to moving Syncplay to Python 3 by any specific date (or whether we would want to support both in parallel by a specific date). Any support people can provide in terms of helping understand and overcome barriers and assessing whether or not there are reasons for urgency or delay would be appreciated.

From the above comments it seems clear there are problems with PySide2 on Python 3. Is the situation of PySide for Python 3 any different to Python 2? In Python 2 my understanding is that PySide1 is fully supported (other than a few OS X quirks) and PySide2 is working if you build it yourself but is not yet stable and accessible enough to be depended upon (especially for Windows).

xNinjaKittyx:
As I was working on it a bit more, I finally was able to find a way to get Pyside2 on Python 3.6 through conda packaging (miniconda or anaconda, I used anaconda) And it seems to work fine. It was hell trying to build it myself (as I don't have much experience in that area), and after 2 hours I found out that conda had a package for python 3.6

[...]

PR #191, opened 2018-07-17
Transition to Python 3.x

albertosottile:
Following the discussion of issue #169, this PR ports all the codebase to Python 3.x on all the platforms. 

A few info and caveats:
- Python 2.7 will be effectively abandoned. After merging, the minimum required version will be 3.4 (required by Twisted).
- **Windows**: to avoid multiple transitions at the same time, in this PR Windows binaries will use Qt 4.8.7, PySide 1.2.4, and py2exe on Python 3.4 (see #185 for details). Transitions to Qt 5.x and pyInstaller (and newer versions of Python) are deferred to the next releases.
- **macOS**: this platform is already on Qt 5.x and binaries will migrate directly to Python 3.6.5, PySide2 5.11.0, and Qt 5.11.1.
- **Linux**: users will be able to run Syncplay with any version of Python newer than 3.4, using Qt 4.x and PySide 1.x  on 3.4, or using Qt 5.x and PySide2 on 3.5+.
- For security reasons, a PR is not allowed to deploy binaries in the CI systems we currently use. If anyone wants to test Syncplay on Python 3.x, I can provide links for binaries, packed by my fork, for Windows and macOS.

Thanks to @Et0h and @xNinjaKittyx for their assistance and support during the port.


buildPy2exe.py
```diff
@@ -439,7 +439,7 @@ def get_nsis_path():
       CreateShortCut "$$SMPROGRAMS\Syncplay\Syncplay.lnk" "$$INSTDIR\Syncplay.exe" ""
       CreateShortCut "$$SMPROGRAMS\Syncplay\Syncplay Server.lnk" "$$INSTDIR\syncplayServer.exe" ""
       CreateShortCut "$$SMPROGRAMS\Syncplay\Uninstall.lnk" "$$INSTDIR\Uninstall.exe" ""
-      WriteINIStr "$$SMPROGRAMS\Syncplay\SyncplayWebsite.url" "InternetShortcut" "URL" "https://syncplay.pl"
+      WriteINIStr "$$SMPROGRAMS\Syncplay\SyncplayWebsite.url" "InternetShortcut" "URL" "http://syncplay.pl"
```
xNinjaKittyx: Is there a reason why this was changed to http?

buildPy2exe.py
```diff
@@ -648,22 +648,22 @@ def getBuildDirContents(self, path):
...
-                if not fileList.has_key(new_root):
+                if new_root not in fileList:
                     fileList[new_root] = []
                 fileList[new_root].append(file_)
         return fileList, totalSize          
     
     def prepareInstallListTemplate(self, fileList):
         create = []
-        for dir_ in fileList.iterkeys():
+        for dir_ in fileList.keys():
```
xNinjaKittyx: In python3 dictionary iteration, it's often better to use `for key in dict` rather than `for key in dict.keys()`. `dict.keys()` returns a list rather than a generator-like object, which `for key in dict` does.

buildPy2exe.py
albertosottile: **EDIT**: fixed. I was looking at the wrong URL, my bad.

buildPy2exe.py
albertosottile: Lot of these changes were done by 2to3. If you want, I can edit this one, but I guess the same issue will appear in multiple other parts of the codebase. If you think these could cause potential problems, I can scroll all the changes and apply the same fix everywhere.

syncplay/protocols.py
```diff
@@ -10,7 +10,7 @@
 
 class JSONCommandProtocol(LineReceiver):
     def handleMessages(self, messages):
-        for message in messages.iteritems():
+        for message in messages.items():
```
xNinjaKittyx: Would this section would be more readable if we used `for command, message in messages.items()` instead?

syncplay/protocols.py
```diff
@@ -75,12 +75,12 @@ def dropWithError(self, error):
         self.drop()
 
     def _extractHelloArguments(self, hello):
-        username = hello["username"] if hello.has_key("username") else None
-        roomName = hello["room"]["name"] if hello.has_key("room") else None
-        version = hello["version"] if hello.has_key("version") else None
-        version = hello["realversion"] if hello.has_key("realversion") else version # Used for 1.2.X compatibility
-        motd = hello["motd"] if hello.has_key("motd") else None
-        features = hello["features"] if hello.has_key("features") else None
+        username = hello["username"] if "username" in hello else None
```
xNinjaKittyx: We could use `hello.get('username', None)` here. and everything else in this section.

buildPy2exe.py
xNinjaKittyx: I see. Yeah there's a lot of places that could be improved with shorter notation, but they're not necessarily "wrong".

I think I'll just ignore them for now, and there can be another PR that cleans up the code.

syncplay/vendor/Qt.py
```diff
@@ -869,12 +869,12 @@ def _wrapinstance(func, ptr, base=None):
 
     """
 
-    assert isinstance(ptr, long), "Argument 'ptr' must be of type <long>"
+    assert isinstance(ptr, int), "Argument 'ptr' must be of type <long>"
```
xNinjaKittyx: Comment should be int and not long right?

Et0h:
@xNinjaKittyx Is everything okay now?

xNinjaKittyx:
@Et0h yeah

PR #192, opened 2018-07-20
Code Syntax Refactoring

xNinjaKittyx:
These are the goals of the PR
- First, it's to apply PEP8 coding standard to most of the code (some things were ignored like bare exceptions or line length under some circumstances) Examples are below
    - `if something ==  True` has been changed to `if something:`. They are not exactly identical, but I *think* most of them are evaluating True/False/None.
    - `if something == False` are not always changed to `if not something`. Ideally, we would want to eliminate `== False` to `is False`. Some code, it's obvious that True/False are the only values available, so I have changed those to an appropriate boolean check.
    - Inline comments formatting, `code  # comment` (2 spaces between code and comment) not `code #comment`
    - All colon usages in dictionaries are adjusted appropriately.
    - kwargs in functions should be `arg=value` not `arg = value`
    - All imports were alphabetized and prioritized in this order: Standard Library, 3rd Party, then Syncplay. Please let me know if any of the top-of-file import orders actually matter.
    - Spaces after commas when giving arguments to a method/function.
    - `if something == None` is changed to `if something is None` and likewise, `if something != None` is changed to `if something is not None`
    - Some code was not spaced in multiples of 4. These have been fixed.

- Secondly, it will hopefully make the code more readable (however, readability can be subjective, so I may be biased on how I decide to space things.)

Issue that I need to fix before merge:
- https://my.mixtape.moe/dbmiob.png - There's a red X next to Room. Not sure how I ended up adding it.

### DIFF
--- a/syncplay/players/vlc.py
+++ b/syncplay/players/vlc.py
@@ -226,11 +226,11 @@ class VlcPlayer(BasePlayer):
             self._paused = bool(value != 'playing') if (value != "no-input" and not self._filechanged) else self._client.getGlobalPaused()
             diff = time.time() - self._lastVLCPositionUpdate if self._lastVLCPositionUpdate else 0
             if (
-                not self._paused and
+                self._paused == False
                 self._position == self._previousPreviousPosition and
                 self._previousPosition == self._position and
                 self._duration > constants.PLAYLIST_LOAD_NEXT_FILE_MINIMUM_LENGTH and
-                (self._duration - self._position) < constants.VLC_EOF_DURATION_THRandD and
+                (self._duration - self._position) < constants.VLC_EOF_DURATION_THRESHOLD and
                 diff > constants.VLC_LATENCY_ERROR_THRESHOLD
             ):
                 self._client.ui.showDebugMessage("Treating 'playing' response as 'paused' due to VLC EOF bug")
--- a/syncplay/players/vlc.py
+++ b/syncplay/players/vlc.py
@@ -1,17 +1,23 @@
-import subprocess
-import re
-import threading
-from syncplay.players.basePlayer import BasePlayer
-from syncplay import constants, utils
+
+import asynchat
+import asyncore
 import os
-import sys
 import random
+import re
 import socket
-import asynchat, asyncore
-import urllib.request, urllib.parse, urllib.error
+import subprocess
+import sys
+import threading
 import time
+import urllib.error
+import urllib.parse
+import urllib.request
+
+from syncplay import constants, utils
 from syncplay.messages import getMessage
+from syncplay.players.basePlayer import BasePlayer
 from syncplay.utils import isBSD, isLinux, isWindows, isMacOS
+
 
 class VlcPlayer(BasePlayer):
     speedSupported = True
@@ -23,8 +29,8 @@ class VlcPlayer(BasePlayer):
     RE_ANSWER = re.compile(constants.VLC_ANSWER_REGEX)
     SLAVE_ARGS = constants.VLC_SLAVE_ARGS
     if isMacOS():
-        SLAVE_ARGS.extend(constants.VLC_SLAVE_MACOS_ARGS)        
-    else:     
+        SLAVE_ARGS.extend(constants.VLC_SLAVE_MACOS_ARGS)
+    else:
         SLAVE_ARGS.extend(constants.VLC_SLAVE_NONMACOS_ARGS)
     vlcport = random.randrange(constants.VLC_MIN_PORT, constants.VLC_MAX_PORT) if (constants.VLC_MIN_PORT < constants.VLC_MAX_PORT) else constants.VLC_MIN_PORT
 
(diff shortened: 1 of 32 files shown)


Et0h:
> if something == True is changed to if something:. Likewise, if something == False it is changed to if not something:
Sometimes we have if not something == False, which is changed to if something

> Issue that I need to fix before merge:
https://my.mixtape.moe/dbmiob.png - There's a red X next to Room. Not sure how I ended up adding it.

Because userReady can be None, checking "elif not userReady" is not functionally equivalent to checking "elif userReady == False". As such, the refactoring changes Syncplay's logic and caused the bug you described.

In some cases values can have three states: True, False and None. The code changes do not recognise this, and incorrectly assume that if something is not True it is False, or if it is not False it is True.

syncplay/ui/gui.py
```diff
@@ -1,28 +1,34 @@
...
+from syncplay import utils, constants, version, revision, release_number
+from syncplay.messages import getMessage
 from syncplay.ui.consoleUI import ConsoleUI
+from syncplay.utils import resourcespath
+from syncplay.utils import isLinux, isWindows, isMacOS
+from syncplay.utils import formatTime, sameFilename, sameFilesize, sameFileduration, RoomPasswordProvider, formatSize, isURL
+from syncplay.vendor import Qt
+from syncplay.vendor.Qt import QtWidgets, QtGui, __binding__, __binding_version__, __qt_version__, IsPySide, IsPySide2
+from syncplay.vendor.Qt.QtCore import Qt, QSettings, QSize, QPoint, QUrl, QLine, QDateTime
+from platform import python_version
```
albertosottile: Is there a reason why this import is not placed before syncplay imports (following your convention)?

syncplay/ui/gui.py
```diff
@@ -87,72 +94,79 @@ def paint(self, itemQPainter, optionQStyleOptionViewItem, indexQModelIndex):
...
+        nameLabel = QtWidgets.QLabel("<center><strong>Syncplay</strong></center>")
+        nameLabel.setFont(QtGui.QFont("Helvetica", 18))
+        linkLabel = QtWidgets.QLabel("<center><a href=\"https://syncplay.pl\">syncplay.pl</a></center>")
+        linkLabel.setOpenExternalLinks(True)
+        versionExtString = version + revision
+        versionLabel = QtWidgets.QLabel(
+            "<p><center>" + getMessage("about-dialog-release").format(versionExtString, release_number) +
+            "<br />Python " + python_version() + " - " + __binding__ + " " + __binding_version__ +
+            " - Qt " + __qt_version__ + "</center></p>")
+        # versionLabel = QtWidgets.QLabel("<p><center>Version 1.5.4 release 62<br />Python 3.4.5 - PySide 1.2.4 - Qt 4.8.7</center></p>")
```
albertosottile: You can remove this commented line.

syncplay/ui/gui.py
Et0h: None that I am aware of. I don't think I ever really paid much attention to the order of imports.

syncplay/ui/gui.py
Et0h: Yeah that makes sense.

syncplay/players/vlc.py
```diff
@@ -212,21 +223,25 @@ def lineReceived(self, line):
...
-            and self._previousPosition == self._position \
-            and self._duration > constants.PLAYLIST_LOAD_NEXT_FILE_MINIMUM_LENGTH \
-            and (self._duration - self._position) < constants.VLC_EOF_DURATION_THRESHOLD \
-            and diff > constants.VLC_LATENCY_ERROR_THRESHOLD:
+            if (
+                not self._paused and
+                self._position == self._previousPreviousPosition and
+                self._previousPosition == self._position and
+                self._duration > constants.PLAYLIST_LOAD_NEXT_FILE_MINIMUM_LENGTH and
+                (self._duration - self._position) < constants.VLC_EOF_DURATION_THRandD and
```
Et0h: Changing to VLC_LATENCY_ERROR_THRESHOLD to VLC_EOF_DURATION_THRandD results in an AttributeError.

syncplay/players/vlc.py
```diff
@@ -212,21 +223,25 @@ def lineReceived(self, line):
...
-            self._paused = bool(value != 'playing') if(value != "no-input" and self._filechanged == False) else self._client.getGlobalPaused()
+            self._paused = bool(value != 'playing') if (value != "no-input" and not self._filechanged) else self._client.getGlobalPaused()
             diff = time.time() - self._lastVLCPositionUpdate if self._lastVLCPositionUpdate else 0
-            if self._paused == False \
-            and self._position == self._previousPreviousPosition \
-            and self._previousPosition == self._position \
-            and self._duration > constants.PLAYLIST_LOAD_NEXT_FILE_MINIMUM_LENGTH \
-            and (self._duration - self._position) < constants.VLC_EOF_DURATION_THRESHOLD \
-            and diff > constants.VLC_LATENCY_ERROR_THRESHOLD:
+            if (
```
Et0h: self._paused can be None, so changing `if self._paused == False `to `not self._paused` might change Syncplay behaviour.

syncplay/players/vlc.py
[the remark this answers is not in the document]
xNinjaKittyx: Thanks for catching that, the diff not being case sensitive is a little annoying...

syncplay/ui/gui.py
xNinjaKittyx: You are correct. Thanks for catching that.

PR #194, opened 2018-07-23
Revert "Code Syntax Refactoring"

albertosottile:
Reverts Syncplay/syncplay#192 due some bugs caused by the refactoring in the tri-state logic: 
wrong file size in the UI, wrong OSD information (readiness and user number).

### DIFF
--- a/appdmg.py
+++ b/appdmg.py
@@ -7,17 +7,15 @@ import os.path
 application = defines.get('app', 'dist/Syncplay.app')
 appname = os.path.basename(application)
 
-
 def icon_from_app(app_path):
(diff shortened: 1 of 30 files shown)


xNinjaKittyx:
Sorry about that. Ill try to see if i can redo without any tri statechanges

albertosottile:
Sorry for this rushed revert, but the release was broken and we had to fix it quickly. This was also our fault, it's clear that we did not test the PR thoroughly enough. Of course, we will welcome a new PR.

Et0h:
I think the safest option is just to not change any logic, only spacing and imports.

PR #195, opened 2018-07-23
Code Syntax Refactoring Try #2

xNinjaKittyx:
This is the same as PR#192, except it reverts almost all `==True` `==False` changes.

All changes should be purely aesthetic except for the following:
- `not key in dict` is changed to `key not in dict`. These are the same, but the latter is standard.
- `x == None` is changed to `x is None` and `x != None` is changed to `x is None`. Unless the object `x` overrides its `__eq__` method to handle it, it should behave exactly the same. Let me know if this is an okay change to make.
- Under `mpv.py` I changed `_fileIsLoaded()` to much simpler logic. Both logic is binary, so it should not affect any functionality.
- Under `ConfigurationGetter.py` I have changed `if str(varToTest).isdigit() == False:` to `if not str(varToTest).isdigit()` since this is a python built-in.

Additional Personal Notes:
- `x == True` ideally in the future should be changed. This syntax forces a check to strict True values such as `True` or `1`, but it makes it so all "truthy" values are False. e.g. `2 == True` is False,
- `x == False` ideally should be changed to `not x and not None` because this allows all 'Falsey' values to evaluate how they're expected to. e.g. `if '' == False` will evaluate as `False` but `if not ''` evaluates True. We add a `and not None` to handle the special None case.


syncplay/utils.py
```diff
@@ -115,52 +126,59 @@ def formatTime(timeInSeconds, weeksAsTitles=True):
         formattedTime = "{0:} (Title {1:.0f})".format(formattedTime, title)
     return formattedTime
 
-def formatSize (bytes, precise=False):
-    if bytes == 0: # E.g. when file size privacy is enabled
+
+def formatSize(num_of_bytes, precise=False):
```
Et0h: You rename bytes 'num_of_bytes' at the top of the def but not in the subsequent code. This results in all sizes being incorrectly reported as '???'.

xNinjaKittyx: Thanks for catching that. I forgot I had made this change. It's to avoid shadowing the bytes keyword.

PR #196, opened 2018-07-25
Allow collection of server statistical data

albertosottile:
Because of the recent increase in the update rate of Syncplay, we would like to have some data of our currently operating user base. 

This PR introduces a new server argument ` --stats-db-file [file]` that creates a SQLite database in [file] and stores some anonymous information in it. 

We only store non-sensitive and non-personal data that are already being transmitted by the clients in the current protocol. Specifically, for every user in the server at the time of stats collection, we store:
- current Unix timestamp
- server port
- user's Syncplay client version
- a numerical index that indicates that the user is in the n-th room of the server, without identifying it
- play status of the n-th room

Data collection is scheduled to happen every hour. Nothing else is acquired nor stored in the snapshot. 

Thanks to SQLite, full concurrency is supported and multiple servers can use the same DB file at the same time. For further safety, a fixed delay is introduced before the first collection, and its amount is based on the server's port number.


syncplay/server.py
```diff
@@ -185,6 +194,17 @@ def setPlaylistIndex(self, watcher, index):
         else:
             watcher.setPlaylistIndex(room.getName(), room.getPlaylistIndex())
 
+    def _connectToStatsDb(self, dbPath):
+        conn = sqlite3.connect(dbPath)
+        c = conn.cursor()
+        c.execute('create table if not exists clients_snapshots (snapshot_time integer, port integer, version string, room_index integer, play_status integer)')
```
Uriziel: would be nice if all the DB stuff:
 - connection
 - schema creating
 - queries

were put into a separate class (service)

syncplay/server.py
```diff
@@ -260,6 +280,16 @@ def moveWatcher(self, watcher, room):
...
         watcher.setFile(watcher.getFile())
 
+    def runClientSnapshot(self, dbHandler, portNumber):
+        snapshotTime = int(time.time())
+        c = dbHandler.cursor()
+        for idx, room in enumerate(self._rooms.values()):
+            playStatus = room.isPlaying()
+            for watcher in room.getWatchers():
+                content = (snapshotTime, int(portNumber), watcher.getVersion(), idx, playStatus, )
+                c.execute("INSERT INTO clients_snapshots VALUES (?, ?, ?, ?, ?)", content)
```
Uriziel: What's the point of saving room state?
Room index is non-deterministic as well and can change within multiple executions on the same dictionary (unless python 3.7).
You'll log multiple users multiple times without any real way to distinguish between them.
Also loging the port makes very litte sense and you should have a separate file for every instance you run (https://www.sqlite.org/faq.html#q5 http://howfuckedismydatabase.com/sqlite/)

syncplay/server.py
```diff
@@ -185,6 +194,17 @@ def setPlaylistIndex(self, watcher, index):
         else:
             watcher.setPlaylistIndex(room.getName(), room.getPlaylistIndex())
 
+    def _connectToStatsDb(self, dbPath):
+        conn = sqlite3.connect(dbPath)
```
Uriziel: maybe you should implement non-blocking db connection:
https://howto.lintel.in/asynchronous-db-operations-twisted/
(haven't tested)

syncplay/server.py
Uriziel: instead of room-idx just store the room-name hash.

daniel-123: You can write to single SQLite database with multiple processes - you just need to handle SQLITE_BUSY and wait for other inserts to finish.

albertosottile: By default, sqlite3 handles SQLITE_BUSY by waiting for 5 seconds and then retrying. That's why a port-based fixed delay was introduced. Having a single SQLite database makes handling and analyses much easier, in my opinion.

syncplay/server.py
```diff
@@ -39,6 +42,12 @@ def __init__(self, password='', motdFilePath=None, isolateRooms=False, salt=None
             self._roomManager = RoomManager()
         else:
             self._roomManager = PublicRoomManager()
+        if statsDbFile is not None:
+            self.statsDbHandle = self._connectToStatsDb(statsDbFile)
+            logDelay = 5*(int(self.port)%10 + 1)
+            reactor.callLater(logDelay, self._scheduleClientSnapshot, self.statsDbHandle, self.port)
```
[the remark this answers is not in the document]
Et0h: The idea is for it to occur at set intervals from being started to provide a snapshot, and to have a delay to avoid multiple servers all writing at exactly the same time if they are started at the same time.

syncplay/server.py
Et0h: Room index can be used to get a sense of the typical size of a session, which can help inform development/design decisions (e.g. default size of 'list of who is watching what' section of client)

syncplay/server.py
albertosottile: We could do that but, quoting the FAQ you linked before, "SQLite allows multiple processes to have the database file open at once". So, I do not think we should rely on twisted for this.

syncplay/server.py
albertosottile: I'll try. Should I keep this new class in server.py or move it to a new file and module? I would advise for the former since all this is server-related only.

syncplay/server.py
daniel-123: If SQLITE_BUSY has potential delay of 5 seconds, won't that risk stalling entire server process for 5 second during *each* iteration of this loop?

albertosottile: My understanding was that only sqlite3.commit locked the database, but probably I was wrong. In any case, I'll try to replace this loop of execute with a list and a single sqlite3.executescript at the end of the loop.

albertosottile: Upon further discussion, @Et0h and I decided to get rid of room_index and play_status for now. The only recorded stats will be the current timestamp, the client's version, and the server port, needed to understand server's load while keeping only one database file.

syncplay/server.py
```diff
@@ -185,6 +192,45 @@ def setPlaylistIndex(self, watcher, index):
         else:
             watcher.setPlaylistIndex(room.getName(), room.getPlaylistIndex())
 
+class StatsRecorder(object):
+    def __init__(self):
+        self._dbPool = None
```
Uriziel: Nitpicking: I'd name this `_connection`

syncplay/server.py
```diff
@@ -185,6 +192,45 @@ def setPlaylistIndex(self, watcher, index):
...
+         
+    def __del__(self):
+        if self._dbPool is not None:
+            self._dbPool.close()
+        
+    def startRecorder(self, dbpath, roomManager, delay):
```
Uriziel: Nitpicking I'd move `dbpath, roomManager` into `__init__`

syncplay/server.py
```diff
@@ -185,6 +192,45 @@ def setPlaylistIndex(self, watcher, index):
...
+        self._roomManagerHandle = roomManager
+        try:
+            self._dbPool = self._initDatabase()
+            reactor.callLater(delay, self._scheduleClientSnapshot)
+        except:
+            self._dbPool = None
+            print("--- Error in initializing the stats database. Server Stats not enabled. ---")
+    
+    def _initDatabase(self):
+        dbpool = adbapi.ConnectionPool("sqlite3", self._dbPath, check_same_thread=False)
```
Uriziel: My idea for separate class was to cover "Single Responsibility Principle" which moves all stuff related to `sqlite` to a separate class, thus making abstraction layer between business code (which is "log this and that data") and app logic (which is "use database to store data").
So something like this:
```
class Database(NameForChange):
    __init__(dbPath): pass
    connect(): pass
   addVersionLog(version, time): pass
    _createSchema(): pass
```

which later on can be extended by adding function like `fetchUser(username)` etc.

Uriziel:
Reposting my last comment as it got squashed by one of the changes (my bad.)

My idea for separate class was to cover "Single Responsibility Principle" which moves all stuff related to sqlite to a separate class, thus making abstraction layer between business code (which is "log this and that data") and app logic (which is "use database to store data").
So something like this:

class Database(NameForChange):
    __init__(dbPath): pass
    connect(): pass
   addVersionLog(version, time): pass
    _createSchema(): pass
which later on can be extended by adding function like fetchUser(username) etc. that all relates to fetching data from data layer (i.e. sqlite)

PR #214, opened 2019-01-10
Fixes #210 and some refactoring

### LINKED ISSUE #210: UnicodeDecodeError
My syncplay server crashed while I was not using it, this was in the error log:
```
builtins.UnicodeDecodeError: 'utf-8' codec can't decode byte 0x89 in position 2: invalid start byte
    line = line.decode('utf-8').strip()
  File "/usr/lib/syncplay/syncplay/protocols.py", line 34, in lineReceived
    why = self.lineReceived(line)
  File "/usr/lib/python3.7/site-packages/twisted/protocols/basic.py", line 572, in dataReceived
    rval = self.protocol.dataReceived(data)
  File "/usr/lib/python3.7/site-packages/twisted/internet/tcp.py", line 249, in _dataReceived
    return self._dataReceived(data)
  File "/usr/lib/python3.7/site-packages/twisted/internet/tcp.py", line 243, in doRead
    why = selectable.doRead()
  File "/usr/lib/python3.7/site-packages/twisted/internet/posixbase.py", line 614, in _doReadOrWrite
--- <exception caught here> ---
    return func(*args,**kw)
  File "/usr/lib/python3.7/site-packages/twisted/python/context.py", line 85, in callWithContext
    return self.currentContext().callWithContext(ctx, func, *args, **kw)
  File "/usr/lib/python3.7/site-packages/twisted/python/context.py", line 122, in callWithContext
    return context.call({ILogContext: newCtx}, func, *args, **kw)
  File "/usr/lib/python3.7/site-packages/twisted/python/log.py", line 86, in callWithContext
    return callWithContext({"system": lp}, func, *args, **kw)
  File "/usr/lib/python3.7/site-packages/twisted/python/log.py", line 103, in callWithLogger
Unhandled Error
```

I guess just move the decode into the try block to fix it?
https://github.com/Syncplay/syncplay/blob/master/syncplay/protocols.py#L34

xNinjaKittyx:
Fixes Bug #210 by not crashing, and spits out a message "Not a utf-8 string"

Took the liberty to generally refactor `protocols.py` in two ways.

1) I noticed one for loop that used the `for key, value in dict.items():`, which I then applied to all for loops where applicable.

2) `dict['key'] if key in dict else None` pattern has been changed to `dict.get('key')`  and if `None` is something else like `0` then you will see `dict.get('key', 0)`

Edit: Would also like to mention that this should improve performance slightly since it'll only do a lookup in the dict once instead of twice.

I also changed handleMessages to use a dict instead of a series of if/elif/else. It takes a more EAFP approach and makes it a little easier to read in my opinion. Let me know if you like or dislike the change.

Tested it working in a Python 3.6 environment (server + 2 clients)


Et0h:
I would need to do extensive testing before allowing a refactor because the last time I allowed one it broke things in subtle way that were hard to detect and track down. I do not currently have time for this at present so I will leave this PR open for now.

Et0h:
Hopefully resolved now by fb9b3ce.

PR #219, opened 2019-02-04
Syncplay Logo Update

catlinman:
Replaced the Windows and MacOS logo files with the new and redesigned logo.


albertosottile:
Travis is failing due to the recent `homebrew` update. I am afraid we will have to abandon the support for macOS 10.11 for the next releases.

PR #226, opened 2019-02-23
Workaround for older twisted versions which don't support raiseMinimumTo.

daniel-123:
Twisted version 16.6.0 and older don't support raiseMinimumTo argument in CertificateOptions. This allows TLS to work on distributions like Debian Stretch (current stable) and Ubuntu 16.04 LTS.

### DIFF
--- a/syncplay/server.py
+++ b/syncplay/server.py
@@ -230,7 +230,7 @@ class SyncFactory(Factory):
                 contextFactory = ssl.CertificateOptions(privateKey=privKeyPySSL, certificate=certifPySSL,
                                                         extraCertChain=chainPySSL, acceptableCiphers=accCiphers,
                                                         raiseMinimumTo=ssl.TLSVersion.TLSv1_2)
-            except:
+            except AttributeError:
                 contextFactory = ssl.CertificateOptions(privateKey=privKeyPySSL, certificate=certifPySSL,
                                                         extraCertChain=chainPySSL, acceptableCiphers=accCiphers,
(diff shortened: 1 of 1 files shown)


syncplay/server.py
```diff
@@ -13,6 +13,7 @@
 try:
     from OpenSSL import crypto
     from twisted.internet import ssl
+    from OpenSSL.SSL import TLSv1_2_METHOD
```
albertosottile: Could you please put the new import right after the first one (and before twisted)?

syncplay/server.py
```diff
@@ -225,9 +226,14 @@ def _allowTLSconnections(self, path):
...
             accCiphers = ssl.AcceptableCiphers.fromOpenSSLCipherString(cipherListString)
 
-            contextFactory = ssl.CertificateOptions(privateKey=privKeyPySSL, certificate=certifPySSL,
-                                                    extraCertChain=chainPySSL, acceptableCiphers=accCiphers,
-                                                    raiseMinimumTo=ssl.TLSVersion.TLSv1_2)
+            try:
+                contextFactory = ssl.CertificateOptions(privateKey=privKeyPySSL, certificate=certifPySSL,
+                                                        extraCertChain=chainPySSL, acceptableCiphers=accCiphers,
+                                                        raiseMinimumTo=ssl.TLSVersion.TLSv1_2)
+            except:
```
albertosottile: Please specify the `AttributeError` exception type here.

Issue #207, opened 2018-11-18
No way to generate optimized bytecode?


alxpettit:
I'm now a co-maintainer of the [syncplay AUR package](https://aur.archlinux.org/packages/syncplay). Ideally, we'd have the installation have `.pyc` files generated and packaged so that they're tracked by `pacman` -- but currently we do not. I'd like to know what the equivalent is of `--optimize=1` for your install toolchain, since you're not using `setup.py`.

Et0h:
I do not have any experience with setup.py, so you might need to find someone with relevant experience to make an appropriate PR.

albertosottile:
In the last few days, I investigated the possibility of using `setuptools` with Syncplay (hence writing a `setup.py` script). Unfortunately, our current project structure does not match the one typically used for python modules or packages, and these differences really stood out in my attempts to adapt `setuptools` to Syncplay, creating a lot of issues.

At this moment, we would like to keep our current folder structure, so we decided to abandon the idea of using also `setuptools` for distributing Syncplay. @alxpettit, going back to your question, a way to generate optimized bytecode for all the python files in Syncplay could be to run 
`python -m compileall .` while being in the repository main folder. Please let us know if this command can be included in the workflow used to pack AUR packages. In the meantime, I am closing this issue.

PR #230, opened 2019-03-25
Add setup.py for setuptools

ObserverOfTime:
This is related to issue #207.

I tried to keep changes to the codebase minimal but I had to edit `syncplay.utils.findWorkingDir` to find the path when installed via `setuptools` since `resources/` is copied into `syncplay/`.
(You could probably use [pkg_resources](https://setuptools.readthedocs.io/en/latest/pkg_resources.html) to load resources and avoid all the hassle.)

Once Syncplay is deployed to [PyPI](https://packaging.python.org/tutorials/packaging-projects/#uploading-the-distribution-archives) (which can be automated by [Travis](https://docs.travis-ci.com/user/deployment/pypi/)), installing it along with its optional dependencies will be as simple as `pip3 install syncplay[gui,tls]`.

**Environment info:**
* OS: `Antergos Linux 5.0.3-arch1-1-ARCH`
* Python: `Python 3.7.2`
* pip: `pip 18.1`
* setuptools: `setuptools 40.8.0`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ dist.7z
 !.travis.yml
 !.appveyor.yml
 __pycache__
+*.egg-info
--- a/setup.py
+++ b/setup.py
@@ -1,0 +1,91 @@
+#!/usr/bin/env python
+
+from os.path import dirname, join
+from setuptools import find_packages, setup
+from setuptools.command import build_py, install_scripts
+from syncplay import projectURL, version
+
+
+def setup_dir(*args):
+    return join(dirname(__file__), *args)
+
+
+def read(fname):
+    with open(setup_dir(fname), 'r') as f:
+        return f.read()
+
+
+class BuildPy(build_py.build_py):
+    def run(self):
+        self.copy_tree(
+            setup_dir('resources'),
+            setup_dir('syncplay', 'resources')
+        )
+        build_py.build_py.run(self)
+
+
+class InstallScripts(install_scripts.install_scripts):
+    def run(self):
+        install_scripts.install_scripts.run(self)
+        scripts = []
+        for script in self.outfiles:
+            old_suff = script[-9:]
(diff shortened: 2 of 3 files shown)


albertosottile:
Thank you for this PR, we really appreciate it. We are currently working on implementing `setuptools` for Syncplay [here](https://github.com/albertosottile/syncplay/commits/setuptools), but I see some differences between this PR and our approach that are definitely worthy of a discussion.

Firstly, you used `requirements*.txt` files in the `setup.py` script. That's something we could do, too. 

Secondly, you found a way to install the scripts without restructuring the package, by subclassing the `InstallScripts` class. While we considered that approach, we think that using entry points is the most natural and future-proof choice, given that they are recommended. So, we restructured some of the code to use entry points, instead of install scripts, as you can see [here](https://github.com/albertosottile/syncplay/commit/6f361ff0d10f5c163af9415584e0e314453b3bc5). 

Thirdly, you subclassed `BuildPy` instead of actually moving the `resources` folder inside the `syncplay` one. While I appreciate that this script does not require to relocate the folder, I am afraid for the future support of this solution and of the possible bugs that it could cause. As an example, this `setup.py` script does not currently copy the `resources` folder when running it with `sdist`. Of course, this and other future bugs can be fixed, but maybe it would just be better to move the `resources` folder now and migrate Syncplay to the recommended file structure for a python package. This is the approach that I took in the fork so far.

In summary, I think we will reject this PR in favor of opening a new one, based on the work done so far in the fork linked above. I sincerely apologize for the time you already invested in working on this. In any case, I strongly encourage to review and contribute to the upcoming PR, since you seem very experienced with `setuptools`.

ObserverOfTime:
Thank you for the response. That's all right, I just did it in an evening as a proof of concept mostly. I wasn't aware that you were already working on it in a fork.

Et0h:
Closed as setuptools is now supported via #231

Issue #229, opened 2019-03-23
Running the client headlessly?


ritiek:
Hi, thanks for the awesome tool!

I'd also like to be able to use the syncplay client on headless machines for only listening to audio. Since, syncplay client pops up the player window (which requires an X server) and due to this one cannot play audio-only files with syncplay client on headless machines. Would something like this be possible?

ritiek:
I got it to work for mpv-player by adding `vo=image` to my `/etc/mpv/mpv.conf` on my headless device! `vo=tct` seems to work as well!

ritiek:
So far I managed to get syncplay dirty work on headless machines, though you still need atleast 1 client with access to X-server. This client (with X-server enabled) will control the music and it would be reflected back on headless devices.

I tried this with only mpv-player. On your headless devices, you need to add these lines in your `/etc/mpv/mpv.conf`:
```
vo=null
no-video
```
and then use syncplay as usual!

----------

I also got this working on Android with Termux. mpv for Termux is built without lua support (lua support was added recently in https://github.com/termux/termux-packages/commit/f1a2af56d2c75aa1ac8af2db33469d5de5b4f583, but it isn't available with `pkg` yet and I don't know much on how to build packages manually for Termux).

Since no lua support, so before compiling syncplay on my Android, I removed this line https://github.com/Syncplay/syncplay/blob/bcf8174b7d4ae723c4bb9788d701a646367a023f/syncplay/players/mpv.py#L44

Also, mpv on Termux does not have access to X11, so you would need to add the above config for mpv here as well (in `/data/data/com.termux/files/usr/etc/mpv/mpv.conf`), so that it works headlessly with Termux.

----------

However, I'd still very much appreciate a command-line only mode for mpv. mpv allows all usual controls such as pause, volume control, seek, etc. through the command-line itself. If this gets implemented, we won't need an X-server enabled device for controlling playback on headless devices. And also, we would be able to control playback on clients via SSH which would be very nice, atleast for me! :D

Et0h:
You could probably hack the Syncplay code to convert chat messages into mpv commands, which means that any client can control things via the chat facility.

You might also be able to enable IPC on mpv to allow for it to be controlled by an IPC script you write: https://github.com/mpv-player/mpv/blob/master/DOCS/man/ipc.rst

PR #231, opened 2019-03-26
Edit folder structure and add setuptools support

albertosottile:
As discussed in issue #207, our current folder structure is different from the typical one for a Python package. Specifically, the `resources` folder should be included in the `syncplay` folder, as all the relevant python script.

This PR includes that file restructuring and also adds a `setup.py` script that relies on it. Entry points are used for both client and server. Requirements are taken directly from the corresponding text files (thanks to PR #230). 

Additionally, the packages needed for TLS are now included in the main requirements, as we do not really want the TLS support to be optional. Also, AppVeyor has been updated to the latest `py2exe` ([link](https://github.com/albertosottile/py2exe/releases/tag/v0.9.3.0)) to enable the embedding of TLS libraries. Incidentally, Windows binaries are also upgraded to Python 3.6.


setup.py
```diff
@@ -0,0 +1,64 @@
...
+import setuptools
+
+from syncplay import projectURL, version as syncplay_version
+
+def read(fname):
+    with open(fname, 'r') as f:
+        return f.read()
+
+installRequirements = read('requirements.txt').splitlines() +\
+                      read('requirements_gui.txt').splitlines()
```
ObserverOfTime:
```suggestion
```

setup.py
```diff
@@ -0,0 +1,64 @@
...
+import distutils.command.install_scripts
+import setuptools
+
+from syncplay import projectURL, version as syncplay_version
+
+def read(fname):
+    with open(fname, 'r') as f:
+        return f.read()
+
+installRequirements = read('requirements.txt').splitlines() +\
```
ObserverOfTime:
```suggestion
```

setup.py
```diff
@@ -0,0 +1,64 @@
...
+    description=' '.join([
+        'Client/server to synchronize media playback',
+        'on mpv/VLC/MPC-HC/MPC-BE on many computers'
+    ]),
+    long_description=read('README.md'),
+    long_description_content_type="text/markdown",
+    url=projectURL,
+    download_url=projectURL + 'download/',
+    packages=setuptools.find_packages(),
+    install_requires=installRequirements,
```
ObserverOfTime:
```suggestion
    install_requires=read('requirements.txt').splitlines(),
    extras_require={'gui': read('requirements_gui.txt').splitlines(),
```

Et0h:
> It might be better to place `requirements_gui.txt` in `extras_require` since some users may not want the GUI (and `PySide2` is several MBs large). You'll just need to inform end users that they'll need `syncplay[gui]` for the full package.

Aside from saving a few megabytes of storage space and speeding up the installation time, are there any circumstances where automatically installing PySide2 would be disruptive, prohibitive or impossible?

Et0h:
@ObserverOfTime
> It might be better to place `requirements_gui.txt` in `extras_require` since some users may not want the GUI (and `PySide2` is several MBs large). You'll just need to inform end users that they'll need `syncplay[gui]` for the full package.

Aside from saving a few megabytes of storage space and speeding up the installation time, are there any circumstances where automatically installing PySide2 would be disruptive, prohibitive or impossible?

ObserverOfTime:
I think #229 qualifies as a use case where the GUI is not needed.

daniel-123:
Personally I think that the vast majority of users do use the GUI and would expect the "software_name" package to deliver that functionality. I would prefer not to even slightly inconvenience them by making them install package with extras. Even if there exist some use cases where such packaging would be more convenient.

What is possibly worth considering is that most cases where one would care about not including superfluous dependencies are headless server. So maybe that should be a completely separate, minimalist package?

I'm not entirely familiar with setuptools ecosystem, but at least with deb/rpm packaging such problem would be usually solved by existence of package syncplay-no-gui that's either required by full syncplay or is mutually exclusive with it.

ObserverOfTime:
I don't think what you're proposing is feasible with `setuptools`.
Additionally, it's not possible to declare default or opt-out requirements.

How about this solution:
```diff
diff --git a/setup.py b/setup.py
index 91b2602..a083b15 100644
--- a/setup.py
+++ b/setup.py
@@ -3,14 +3,16 @@
 import distutils.command.install_scripts
 import setuptools
 
+from os import environ as env
 from syncplay import projectURL, version as syncplay_version
 
 def read(fname):
     with open(fname, 'r') as f:
         return f.read()
 
-installRequirements = read('requirements.txt').splitlines() +\
-                      read('requirements_gui.txt').splitlines()
+installRequirements = read('requirements.txt').splitlines()
+if env.get('SYNCPLAY_NO_GUI', '0') != '1':
+    installRequirements += read('requirements_gui.txt').splitlines()
 
 setuptools.setup(
     name="syncplay",
```

This way, regular users will not be affected and anyone who doesn't want the GUI can just set `SYNCPLAY_NO_GUI=1` prior to the installation and update.

albertosottile:
Will that work for wheels too? I wonder if the sake of not having `PySide2` installed on a system is worth amending in this way the `setup.py` script, introducing a new environment variable, documenting it somewhere, maintaining the whole thing...

Et0h:
@ObserverOfTime Thanks for your suggestion regarding how we could use environmental variables as a way of supporting a 'no GUI required' installation option without making it the default - we will look into it further if we ever decide to pursue PyPi support.

PR #239, opened 2019-06-21
Changing the file now automatically resets the time to 0:00

djava:
It's just always annoyed me when I change the file and then it starts at the end, so I fixed it.

### DIFF
--- a/syncplay/client.py
+++ b/syncplay/client.py
@@ -474,6 +474,7 @@ class SyncplayClient(object):
         self.userlist.currentUser.setFile(filename, duration, size, path)
         self.sendFile()
         self.playlist.changeToPlaylistIndexFromFilename(filename)
+        self.setPosition(0)
 
     def setTrustedDomains(self, newTrustedDomains):
         from syncplay.ui.ConfigurationGetter import ConfigurationGetter


Et0h:
Thanks for raising this issue and trying to fix it.

I don't mind the behaviour of position resetting when someone double clicks for a different file in the playlist, but for it to reset every time updateFile is called seems excessive.

I've not double checked, but I think that the updateFile function can be also called due to:
- late duration information being provided
- user switching from an incomplete version of a file to the complete version of a file (e.g. if they are downloading it as they are playing)
- when they join an existing channel where something is being played.

In all of those use cases I think it would be inappropriate for the file position to be reset.

My suspicion is that the fix might involve changing the player code for VLC and MPC-HC to ensure that the resetPosition value in the openFile function results in the position being set to 0. I think mpv's behaviour in this regard is the correct behaviour: https://github.com/Syncplay/syncplay/blob/master/syncplay/players/mpv.py#L269 but if you look at https://github.com/Syncplay/syncplay/blob/master/syncplay/players/vlc.py#L168 and https://github.com/Syncplay/syncplay/blob/master/syncplay/players/mpc.py#L407 the resetPosition value seems to be ignored.

djava:
Oh, ok.  I really hadn’t looked super closely through the source code to see where updateFile() was called, so I was just looking for a quick fix.  If I have time later, I’ll try to redo it better.

djava:
Do you know where double clicking something on the shared playlist is handled? I looked through the GUI code a little bit last night, but after not seeing anything immediately, I gave up and put the setPosition call in updateFile.  Wherever that is is probably the easiest place to put in a setPosition(0) call to reset the position.

Et0h:
> Do you know where double clicking something on the shared playlist is handled? I looked through the GUI code a little bit last night, but after not seeing anything immediately, I gave up and put the setPosition call in updateFile. Wherever that is is probably the easiest place to put in a setPosition(0) call to reset the position.

As per https://github.com/Syncplay/syncplay/blob/master/syncplay/ui/gui.py#L811 playlistItemClicked calls  self._syncplayClient._player.openFile with resetPosition=True

There is also https://github.com/Syncplay/syncplay/blob/master/syncplay/ui/gui.py#L693 and https://github.com/Syncplay/syncplay/blob/master/syncplay/ui/gui.py#L840 where you can right click on a file and select an option to open or switch to a file or stream. These call self.openFile (resetPosition=True) which in turn calls self._syncplayClient._player.openFile with resetPosition=True.

As per https://github.com/Syncplay/syncplay/blob/master/syncplay/ui/gui.py#L983 if you open a file via browseMediapath then it calls self._syncplayClient._player.openFile(fileName) without setting resetPosition to True.

As per https://github.com/Syncplay/syncplay/blob/master/syncplay/ui/gui.py#L1790 if you use a dropEvent to drag a file into Syncplay whether or not it sets resetPosition to true will depend upon what keys you are holding down as you drag. As per https://syncplay.pl/guide/client/ if you hold down shift (make it a MoveAction) it is supposed to reset the position but otherwise it won't.

Et0h:
> Do you know where double clicking something on the shared playlist is handled? I looked through the GUI code a little bit last night, but after not seeing anything immediately, I gave up and put the setPosition call in updateFile. Wherever that is is probably the easiest place to put in a setPosition(0) call to reset the position.

In my understanding the reset should be handled the by the player which gets given the resetPosition command as part of an openFile command. In reality, the player doesn't need to seek to 0 because when it loads a file it should start there, it just needs to store the position as 0 internally so when it tells the player the position it tells it 0 if the position has been reset.

As per https://github.com/Syncplay/syncplay/blob/master/syncplay/ui/gui.py#L811 playlistItemClicked calls  self._syncplayClient._player.openFile with resetPosition=True

There is also https://github.com/Syncplay/syncplay/blob/master/syncplay/ui/gui.py#L693 and https://github.com/Syncplay/syncplay/blob/master/syncplay/ui/gui.py#L840 where you can right click on a file and select an option to open or switch to a file or stream. These call self.openFile (resetPosition=True) which in turn calls self._syncplayClient._player.openFile with resetPosition=True.

As per https://github.com/Syncplay/syncplay/blob/master/syncplay/ui/gui.py#L983 if you open a file via browseMediapath then it calls self._syncplayClient._player.openFile(fileName) without setting resetPosition to True.

As per https://github.com/Syncplay/syncplay/blob/master/syncplay/ui/gui.py#L1790 if you use a dropEvent to drag a file into Syncplay whether or not it sets resetPosition to true will depend upon what keys you are holding down as you drag. As per https://syncplay.pl/guide/client/ if you hold down shift (make it a MoveAction) it is supposed to reset the position but otherwise it won't.

If you want to handle the resetPosition via client.py's openFile then you would need to ensure that Syncplay never calls the player's openFile directly with resetPosition=true.

allbombson:
Nice this always annoyed me

djava:
Totally forgot about this - I tried to add a handler for reset position in
vlc.py, but weird things happened every time you used it (vlc and Syncplay
just being totally out of sync, other weird behavior), and after a couple
more attempts to fix it, I gave up.

soredake:
Any progress on this?

Et0h:
@djava @soredake @allbombson The issue addressed by this PR is hopefully resolved by 7d1fda5521afd9687c91a07c5c10b927d1d2bfc7 which is available as part of https://bintray.com/syncplay/Syncplay/Syncplay/v1.6.5#files - please test it and let me know.

Et0h:
This issue is hopefully resolved by 7d1fda5521afd9687c91a07c5c10b927d1d2bfc7 which is available as part of https://bintray.com/syncplay/Syncplay/Syncplay/v1.6.5#files - please test it and let me know.

Issue #242, opened 2019-06-22
[Feature request] TLS with fullchain.pem file


Tremolo4:
The TLS functionality on the server apparently requires a `cert.pem`, `chain.pem` and `privkey.pem` file. However my hosting provider (www.uberspace.de) only provides Let's Encrypt certificates in the form of `fullchain.pem` and `privkey.pem`.

Essentially not a big deal since the actual certificates are the same: currently I'm using a small script to split the `fullchain.pem` file -- first cert is `cert.pem`, any additional certs are `chain.pem`.

I was wondering whether syncplay could be taught to read the certificates out of a single fullchain.pem file as well, so I don't have to bother with splitting it.

Thanks for reading :)

albertosottile:
Believe me, we tried to get rid of the extra `chain.pem` file but, as far as we know, that is currently not possible with Twisted. See [their documentation](https://twistedmatrix.com/documents/current/api/twisted.internet.ssl.CertificateOptions.html) and also [this post](https://community.letsencrypt.org/t/pyopenssl-twisted-servers-certificate-chain-is-incomplete/18591). You can also find more information about our implementation of TLS in the comments of issue #217.

If things will change in newer releases of Twisted, I would be more than happy to amend this feature.

Tremolo4:
The insufficient APIs there are indeed annoying. But I've managed to come up with something that works!  #243

PR #243, opened 2019-06-23
Better TLS certificate chain support

Tremolo4:
This PR adds support for `fullchain.pem` files, i.e., where the main/leaf server certificate is included in the chain file.
Additionally, `chain.pem` or `fullchain.pem` files with more than one intermediate certificate are now supported.

The PEM splitting function might look a bit hacky, but it should be perfectly safe and doesn't require any dependencies.

I've changed the last-edit-check to use the most recent timestamp out of all the four `.pem` files.

Hope this is useful! (it is for me ;))

### DIFF
--- a/syncplay/protocols.py
+++ b/syncplay/protocols.py
@@ -688,7 +688,7 @@ class SyncServerProtocol(JSONCommandProtocol):
         if "send" in inquiry:
             if not self.isLogged() and self._factory.serverAcceptsTLS:
                 lastEditCertTime = self._factory.checkLastEditCertTime()
-                if lastEditCertTime is not None and lastEditCertTime != self._factory.lastEditCertTime:
+                if lastEditCertTime and lastEditCertTime != self._factory.lastEditCertTime:
                     self._factory.updateTLSContextFactory()
                 if self._factory.options is not None:
                     self.sendTLS({"startTLS": "true"})
--- a/syncplay/server.py
+++ b/syncplay/server.py
@@ -209,6 +209,30 @@ class SyncFactory(Factory):
         else:
             watcher.setPlaylistIndex(room.getName(), room.getPlaylistIndex())
 
+    def _splitPEMCerts(self, pem):
+        begin = '-----BEGIN CERTIFICATE-----'
+        end = '-----END CERTIFICATE-----'
+        certs = []
+        while True:
+            # skip until next begin line if exists
+            parts = pem.split(begin, 1)
+            if len(parts) > 1:
+                pem = begin + parts[1]
+                # skip until next end line if exists
+                parts = pem.split('-----END CERTIFICATE-----', 1)
+                if len(parts) > 1:
+                    # found a cert, add it
+                    certs.append(parts[0] + end) 
+                    # remainder of file
+                    pem = parts[1] 
+                else:
+                    # missing end line ... invalid file
+                    break
+            else:
+                # no more certs, done
+                break
+        return certs
+
     def _allowTLSconnections(self, path):
         try:
             privKey = open(path+'/privkey.pem', 'rt').read()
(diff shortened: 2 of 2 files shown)


Et0h:
Hi @Tremolo4, your code looks like a clever way of circumventing the limitations of Twisted. However, when it comes to encryption our philosophy is to avoid custom implementations as much as possible.

We do not have the resources and expertise to review or maintain a custom implementation. For all we know there was a reason why Twisted did things the way they did, or the specific implementation might work fine with the current version of Twisted but might be problematic in a future version.

As such, our view is that if the issue raised in #242 is to be resolved then it would have to be resolved upstream by Twisted. You can try working with the Twisted team directly on this, but it is up to them which development efforts they wish to prioritise.

Until it is fixed upstream our position is that the matter is closed. However, we recognise that some users will experience the same issue as you, so if you can provide guidance on how people can manually split the fullchain.pem file then we would be happy to add it to the guide at https://github.com/Syncplay/syncplay/wiki/TLS-support

Tremolo4:
That's fair, thanks for the response. In general I completely agree with your philosophy about leaving encryption to competent implementations as much as possible.

However, I personally think that my PR doesn't interfere with that philosophy, since we're still fully relying on library code to guarantee the security:
* Actually parsing the certificates is done by PySSL/OpenSSL: it will throw an error if, for example, we pass it an invalid certificate.
* If the server is missing a certificate, for example due to a programming mistake on my side, the client will refuse connecting to the server.

But I guess it's a matter of interpretation and you have more leverage here :D

Just one more point: Currently, it is impossible to use TLS with syncplay if your certificate requires more than one intermediate certificate (i.e., `chain.pem` contains multiple certs). And that is due to syncplay code only: The twisted API does provide that functionality (list of certs), but syncplay never passes more than one in the list. (The problem is that PySSL's `load_certificate` only parses one certificate at a time).

Anyway, thanks a lot for your work on syncplay and thanks for taking the time to discuss this!
Feel free to just ignore this if the decision is final.

albertosottile:
> Just one more point: Currently, it is impossible to use TLS with syncplay if your certificate requires more than one intermediate certificate (i.e., chain.pem contains multiple certs). 

Thanks for raising this issue. While we suspected this condition, were not fully aware of it and we will try to address this in the future.

> And that is due to syncplay code only: The twisted API does provide that functionality (list of certs), but syncplay never passes more than one in the list. (The problem is that PySSL's load_certificate only parses one certificate at a time).

I tend to disagree with the first statement. As you said, the issue is that `pyOpenSSL.crypto.load_certificate` handles only the first certificate in a passed `.pem` file, as reported [here](https://github.com/pyca/pyopenssl/issues/462). Hence, I do not see this as an issue caused by syncplay code only, since we are still delegating certificate handling to dedicated libraries. 

A potential solution to the latter issue and to the issue originally described by this PR could be to use another external library (e.g. `pem` - [GitHub](https://github.com/hynek/pem)) to handle the certificates instead of pyOpenSSL directly. 

While this approach seems appealing in principle, I remember well that implementing TLS for Syncplay required a lot of trial and error, a lot of failed attempts, and a lot of compromises between supported features and reliability. Hence, for the moment I would tend to maintain the current TLS code as much as possible, while perhaps experimenting privately with `pem`, then upgrade to a new logic/library only once that's been extensively tested.

Tremolo4:
> I tend to disagree with the first statement. As you said, the issue is that pyOpenSSL.crypto.load_certificate handles only the first certificate in a passed .pem file, as reported here. Hence, I do not see this as an issue caused by syncplay code only, since we are still delegating certificate handling to dedicated libraries.

You're completely right, I stand corrected. Let's say the two APIs (PySSL and twisted) are not fully compatible / are incomplete for the desired use case.

When looking into how to fix this situation, I saw hynek's `pem` library as well. But I thought it would be overkill to use it, because the only thing we ever need out of the `chain.pem` files are the `-----BEGIN CERTIFICATE-----`...`-----END CERTIFICATE-----` blocks -- as implemented in the PR. I was faced with deciding between including another external library or writing 10-ish lines of code and opted for the latter :P

albertosottile:
@Tremolo4 Syncplay now ships `pem`, as we used it to mitigate #554. Feel free to reconsider this PR basing on this library, if you are still interested in this feature.

Issue #248, opened 2019-08-02
TLS more.....


fmohican:
It's possible to add support for more format like PKCS#7, DER, PKCS#12, and support for self-signed certificate (single cert. or CA)

Thanks!

daniel-123:
While adding support for multiple certificate formats would be feasible, I don't really see a good enough justification for it:
- PEM we support is the same format you get certificates from Let's Encrypt which is by far the most easily available source of certificates signed by roots that are trusted by most software.
- Supporting one common format is basically the industry standard among web servers etc.
- You can easily convert between formats using external tools.

With regards to the second part about self-signed "single cert. or CA": it's not completely clear to me what you mean. I assume you mean two scenarios:
- A typical self-signed certificate
- Complete certificate chain, but signed with your own CA.

Second scenario of those two will already work. Only extra step is that clients also need to add your CA to their trusted CA bundle.

With regards to first scenario with self-signed cert I have similar opinion as with certificate formats - it is feasible, but I see little justification:
- You can easily get free domain validated certificates from Let's Encrypt.
- You can easily get free domain names from likes of duckdns.org.
- In terms of provided features self-signed certificate is very similar to situation with rolling your own CA, which is already possible to use.
- It would probably imply extra work with regards to the UI (I assume you don't want to just direct the user to manually add the cert to their trusted CA bundle).

fmohican:
>
> * Complete certificate chain, but signed with your own CA.
> 
> Second scenario of those two will already work. Only extra step is that clients also need to add your CA to their trusted CA bundle.
> 

That will work in my case, but will be nice if we can just have TSL (or only TSL), without 'server validation', just to crypt packages, also i already have my own _free_ domain (.ml/.ga/.tk), with Let's Encrypt.
However Let's Encrypt issues format PEM format but with extension ".crt" & ".key" (in term of format ins't any difference), but syncplay work only with `.pem` extension.

Will be nice if server had a config file and a GUI, to set all thing, after just register as "services" (on windows).

Thanks for quick answer. That help me a lot, i was unsure about "own CA scenario". Thanks for make crystal clear.

daniel-123:
>However Let's Encrypt issues format PEM format but with extension ".crt" & ".key" 

[...]

PR #252, opened 2019-08-17
Use ca configuration if defined

kidburglar:
It let the user push his own CA to syncplay by override certPath.

I saw that certifi.where() use his own cacert bundle, it can be usefull to use another one. It also fixed the second scenario of #248 when we use a self-signed certificate with a custom CA.

### DIFF
--- a/syncplay/client.py
+++ b/syncplay/client.py
@@ -129,6 +129,11 @@ class SyncplayClient(object):
 
         self._serverSupportsTLS = True
 
+        """Override certPath if defined in the configuration"""
+        if config['ca']:
+            certPath = config['ca']
+            os.environ['SSL_CERT_FILE'] = certPath
+
         if constants.LIST_RELATIVE_CONFIGS and 'loadedRelativePaths' in self._config and self._config['loadedRelativePaths']:
             paths = "; ".join(self._config['loadedRelativePaths'])
             self.ui.showMessage(getMessage("relative-config-notification").format(paths), noPlayer=True, noTimestamp=True)
--- a/syncplay/ui/ConfigurationGetter.py
+++ b/syncplay/ui/ConfigurationGetter.py
@@ -92,7 +92,8 @@ class ConfigurationGetter(object):
             "notificationTimeout": 3,
             "alertTimeout": 5,
             "chatTimeout": 7,
-            "publicServers": []
+            "publicServers": [],
+            "ca": None
         }
 
         self._defaultConfig = self._config.copy()
@@ -191,7 +192,7 @@ class ConfigurationGetter(object):
                 "autoplayInitialState", "mediaSearchDirectories",
                 "sharedPlaylistEnabled", "loopAtEndOfPlaylist",
                 "loopSingleFiles",
-                "onlySwitchToTrustedDomains", "trustedDomains", "publicServers"],
+                "onlySwitchToTrustedDomains", "trustedDomains", "publicServers", "ca"],
             "gui": [
                 "showOSD", "showOSDWarnings", "showSlowdownOSD",
                 "showDifferentRoomOSD", "showSameRoomOSD",


kidburglar:
Hello @daniel-123 I think certifi use his own bundled CA and not the system one.
If I'm right it mean there is no way to properly manage your own CA. That is a big trouble because it result to manage 2 CA bundled and the need to do it properly.

Let me know if I'm wrong about certifi CA bundle.
Cheers.

daniel-123:
Right now Syncplay uses the CA bundle that's returned by certifi.where()

At least on Debian and Fedora that returns path to system CA bundle, I'm not sure about other distributions. On Windows and Mac it indeed uses the CA bundle delivered by certifi.

Back to my question - I still do not see any reasonably typical use case for manually switching to a different CA bundle. I get that this is somewhat common functionality in enterprise software, but Syncplay isn't really aiming at being that.

kidburglar:
Maybe it's related if it's installed by the system or by --user flag but in my case it didn't use the system one, output for python2 and python3.

```
>>> import certifi.where()
>>> import certifi
>>> certifi.where()
'/home/user/.local/lib/python2.7/site-packages/certifi/cacert.pem'
```

```
>>> import certifi
>>> certifi.where()
'/home/user/.local/lib/python3.7/site-packages/certifi-2019.6.16-py3.7.egg/certifi/cacert.pem'
```

Simply if we not want to make a register though letsencrypt. With the output above it could lead to another issue if by bad luck we use a certificate that the CA is not recognized by the CA bundle.

daniel-123:
The problem I have with that use case, is that it requires you to ensure that your custom certificate root needs to be added to each and every client you want to have connected to your server. This is done on regular basis in enterprise environment by using Group Policies on Windows, but I don't think it makes much sense outside of that specific context.

kidburglar:
I just checked the project python-certifi and they clearly show that it use his own CA.
It's something that can maybe work pretty well on Windows but on Linux there is a reason that we have a system CA. It's more easy to be manage but also for security concerns it's more easier.

Based on what I see when I use SSL for websites, we always receive the CA with our certificates. I think that the main reason it is given, it's because it's always possible that our bundle not have that CA.

daniel-123:
After some poking around I see why certifi.where() it's different for you. On my system I use certifi from repositories which differs from upstream and actually points towards system CA. Is this an option for you?

>it's always possible that our bundle not have that CA.

I'm not sure if I understand. Standard CA bundles (like Mozilla one used by certifi, or default one in Debian) are all pretty much the same.

Only reason to add a CA to them is if you have your own one. And in that case certificates signed by that CA will not show up as trusted in web browser unless that CA was specifically added to the bundle given browser uses (Chrome and Edge use system CA, but Firefox does not).

>It's something that can maybe work pretty well on Windows but on Linux there is a reason that we have a system CA.

I know this is somewhat offtopic nitpicking, but it's actually the other way around. Windows does have a single standard CA bundle while Linux does not. Linux distributions usually have some semblance of system-wide CA, but you cannot actually rely on it being there.

kidburglar:
> After some poking around I see why certifi.where() it's different for you. On my system I use certifi from repositories which differs from upstream and actually points towards system CA. Is this an option for you?

I confirm that using the repositories make it use the system bundle. For me it's ok then.
> 
> > it's always possible that our bundle not have that CA.
> 
> I'm not sure if I understand. Standard CA bundles (like Mozilla one used by certifi, or default one in Debian) are all pretty much the same.
> 
> Only reason to add a CA to them is if you have your own one. And in that case certificates signed by that CA will not show up as trusted in web browser unless that CA was specifically added to the bundle given browser uses (Chrome and Edge use system CA, but Firefox does not).
> 
> > It's something that can maybe work pretty well on Windows but on Linux there is a reason that we have a system CA.
> 
> I know this is somewhat offtopic nitpicking, but it's actually the other way around. Windows does have a single standard CA bundle while Linux does not. Linux distributions usually have some semblance of system-wide CA, but you cannot actually rely on it being there.

I would still advice a little note about which bundled is used in the documentation so if someone need to add his own CA there an he not use the system bundle he still can do it on his own.

PR #253, opened 2019-08-17
Add error message if SAN doesn't match hostname

kidburglar:
When I was trying to make syncplay working with a self-signed certificate I had a reconnection loop.

The issue was related to the SAN (Subject Alternative Name) that was not correct in my certificate.
I think it can be usefull to detect this error.

### DIFF
--- a/syncplay/messages_de.py
+++ b/syncplay/messages_de.py
@@ -320,6 +320,7 @@ de = {
     "startTLS-initiated": "Attempting secure connection",
     "startTLS-secure-connection-ok": "Secure connection established ({})",
     "startTLS-server-certificate-invalid": 'Secure connection failed. The server uses an invalid security certificate. This communication could be intercepted by a third party. For further details and troubleshooting see <a href="https://syncplay.pl/trouble">here</a>.',
+    "startTLS-server-certificate-invalid-DNS-ID": "Syncplay does not trust this server because it uses a certificate that is not valid for its hostname.",
     "startTLS-not-supported-client": "This client does not support TLS",
     "startTLS-not-supported-server": "This server does not support TLS",
 
--- a/syncplay/messages_en.py
+++ b/syncplay/messages_en.py
@@ -1,4 +1,4 @@
-# coding:utf8
+startTLS-server-certificate-invalid-DNS-ID# coding:utf8
 
 """English dictionary"""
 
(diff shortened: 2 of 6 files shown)


kidburglar:
Hello @albertosottile @Et0h the stubs is to copy/paste the translation in the other langages ?
If it's that I can to it and provide an update so it can be merged.

Et0h:
> Hello @albertosottile @Et0h the stubs is to copy/paste the translation in the other langages ?
> If it's that I can to it and provide an update so it can be merged.

You copy it over and then add '# TODO: Translate' as comment. Then translators can just look for the TODOs to find out what needs translating.

daniel-123:
I'm not against this change, but if there is a more specific and user friendly message about TLS failure it should be worded in way that's more understandable - since I don't expect most people (or even most power users) to know X.509 field names and their purposes by heart.

Not to reinvent the wheel I'd think wording more similar to what browsers use to be more appropriate. For example Firefox in such situation will show the following in advanced description of problem:

> Websites prove their identity via certificates. Firefox does not trust this site because it uses a certificate that is not valid for wrong.host.badssl.com. The certificate is only valid for the following names: *.badssl.com, badssl.com

kidburglar:
@daniel-123 

I'm open to any wording, at first it was to avoid an generic error when it can be more precise.
But if someone can tell me which text I need to use I will do the replacement (or if someone want change it directly in the PR it's good for me also)

Et0h:
@kidburglar If you want to propose wording and update your pull request soon in line with @daniel-123's comments then this can hopefully be included in the release candidate of Syncplay v1.6.5.

kidburglar:
Hello @Et0h 

Like I said before to @daniel-123 I'm open to any wording. There is just need to tell me which wording using and I will make the changes (if it's faster to make the changes directly without that I make it, it's good for me too).

daniel-123:
I would go with minimal modification to the previously mentioned message, so:
`Syncplay does not trust this server because it uses a certificate that is not valid for its hostname.`

Ideally it would spell out both the current hostname it sees and what's valid in certificate it received, but I guess that would be more complexity than this is worth.

kidburglar:
Hello @daniel-123 @Et0h  I have change the message and I add it to the other languages with a "    # TODO: Translate" comment.

PR #275, opened 2020-03-14
Fixes for OSX deploy issues

palfrey:
Deploys appear to be failing due to an issue with Travis installing it's dpl gem. Others (https://travis-ci.community/t/ruby-openssl-python-deployment-fails-on-osx-image/6753) have had the same problem, and this is a workaround from there that should probably work but is hard to check until it's on master.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,11 +7,6 @@ jobs:
     - language: objective-c
       os: osx
       osx_image: xcode9.2
-      addons:
-        homebrew:
-          packages:
-          - python
-          update: true
     - language: minimal
       dist: xenial
       os: linux
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,20 +1,21 @@
-matrix:
+# These two are defaults, which get overriden by the jobs matrix
+language: minimal
+os: linux
+
+jobs:
   include:
     - language: objective-c
+      os: osx
       osx_image: xcode8.3
     - language: bash
-      sudo: required
       dist: xenial
+      os: linux
       env: BUILD_DESTINATION=snapcraft
     - language: python
-      sudo: required
+      os: linux
       dist: xenial
       python: 3.6
       env: BUILD_DESTINATION=appimage
-
-branches:
-  only:
-  - master
 
 script:
 - if [ "$TRAVIS_OS_NAME" == "osx" ]; then python3 buildPy2app.py py2app ; fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ jobs:
   include:
     - language: objective-c
       os: osx
-      osx_image: xcode8.3
+      osx_image: xcode9.2
     - language: minimal
       dist: xenial
       os: linux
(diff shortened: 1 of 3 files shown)


palfrey:
The homebrew change is to make the Travis build work again, as without that for some reason `python3 buildPy2app.py py2app` fails to find `python3` (also it makes the build a lot faster!)

palfrey:
> The homebrew change is to make the Travis build work again, as without that for some reason `python3 buildPy2app.py py2app` fails to find `python3` (also it makes the build a lot faster!)

I think that'll actually mean we don't need the workaround anymore actually, as we're not upgrading openssl anymore.

albertosottile:
Thanks for your contribution. The `brew update` step has been removed and added several times to fix issues with the Travis build... Let's try without it, for a while. 

I have two questions though: 
- Why is the Travis build for this PR not shown on GitHub?
- From the build log ([get from Travis](https://travis-ci.org/github/Syncplay/syncplay/jobs/662405988#L748)) it seems that the build took place on 10.13, but it should use 10.12... Why did the build not run on Sierra, despite the image was correctly configured?  Also, it seems the OS is not really recognized in the Travis interface. 

I have no idea on how to fix these two issues. I would try to put `os: osx` in the macOS part of our matrix, as apparently that is required now (https://docs.travis-ci.com/user/reference/osx/).

palfrey:
>     * Why is the Travis build for this PR not shown on GitHub?

Probably because of the "branch only master" filtering I've now removed.

>     * From the build log ([get from Travis](https://travis-ci.org/github/Syncplay/syncplay/jobs/662405988#L748)) it seems that the build took place on 10.13, but it should use 10.12... Why did the build not run on Sierra, despite the image was correctly configured?  Also, it seems the OS is not really recognized in the Travis interface.
> 
> I have no idea on how to fix these two issues. I would try to put `os: osx` in the macOS part of 
our matrix, as apparently that is required now (https://docs.travis-ci.com/user/reference/osx/).

Looks like this was due to the lack of the `os:osx` bit as you thought.

palfrey:
> Thanks for your contribution. The `brew update` step has been removed and added several times to fix issues with the Travis build... Let's try without it, for a while.

Turns out once you lock it back down to 10.12, it's needed! Will keep prodding away at this and @ you once I've got a proper fix.

palfrey:
@albertosottile Right, I've managed to get it building properly again, and while using specific locked down URLs to Homebrew installs so we a) don't need an explicit upgrade b) have lower odds of breaking in the future when Homebrew updates things again...

albertosottile:
This is great, thanks for going through the hassle of finding the URLs for the right homebrew formulas. Thank you for your contribution.

palfrey:
> This is great, thanks for going through the hassle of finding the URLs for the right homebrew formulas. Thank you for your contribution.

Annoyingly, it turns out that the original fix is no longer needed (and was fiddly to test on non-master, so I hadn't checked before). #279 I think resets this and should now correct the problem...

PR #277, opened 2020-03-19
Fix macOS environment variable patchup:

### LINKED ISSUE #276: PYTHONPATH fix on macOS actually breaks it depending on configuration
```
commit c07206c18992c1dca401b30a01b9f0fe54a71df5
Author: Alberto Sottile <alby128@gmail.com>
Date:   Thu Mar 14 15:45:55 2019 +0100

    Run mpv subprocess with a system PYTHONPATH env on macOS.
    
    youtube-dl relies on system python to run on macOS, but system
    python cannot be executed in a subprocess created from the frozen
    python3 executable. This makes youtube-dl unusable from frozen
    versions of Syncplay on macOS. So, the environment of the mpv
    subprocess is modified with system PYTHONPATH, allowing the
    execution of /usr/bin/python (and hence, of youtube-dl)
    from within the subprocess in frozen executables.
```

This sets `PYTHONPATH` to a a value determined by running `/usr/bin/python` and extracting `sys.path`, as well as hardcoding `PATH` to `/usr/bin:/usr/local/bin`.

It seems that Homebrew does install youtube-dl using `/usr/bin/python`.  But youtube-dl can be installed with any installation of Python, e.g. using `pip install youtube-dl` or by cloning the repository and running `python setup.py install`.  For that matter, it can also be installed at any path; I'd expect Syncplay to respect the existing `PATH` variable.

In my case, I have youtube-dl installed as `/usr/local/bin/youtube-dl`, so Syncplay does see it even with the hardcoded `PATH`.  But the shebang points to `/usr/local/bin/python3`, the copy of Python I installed it with.  When mpv invokes it with `PYTHONPATH` containing a bunch of paths from the system Python 2.7, it naturally fails to start.

It shouldn't be necessary to set `PYTHONPATH` at all; after all, it's not like you normally have to manually set `PYTHONPATH` before running `youtube-dl`.  Instead, I believe the original problem this commit was addressing arose from the fact that Syncplay's frozen Python 3 itself sets `PYTHONPATH`, and this was inadvertently inherited by `youtube-dl`.  A better approach is to simply unset `PYTHONPATH` before running `youtube-dl`.  (Or for maximum correctness, figure out what the `PYTHONPATH` environment variable was originally set to before frozen Python overwrote it, but that may be difficult.)

As for `PATH`, it doesn't seem like frozen Python modifies it, but I guess the desire is to find youtube-dl in /usr/local/bin even though macOS normally runs apps with `PATH` set to `/usr/bin:/bin:/usr/sbin:/sbin` [(though it's alterable)](https://stackoverflow.com/questions/460835/how-to-set-the-path-as-used-by-applications-in-os-x).  The goal makes sense, but some users may use other paths, especially (but not only) when *not* using frozen Python – I normally run Syncplay from terminal.  Instead of replacing the entire `PATH`, Syncplay should probably just add `/usr/local` to the end of `PATH`, so the user's paths take precedence.  Also, `PATH` alteration should probably only happen at all when running under frozen Python; the current code enables it whenever `isMacOS()` returns true.  (I normally run Syncplay directly from a Git checkout using `python3 syncplayClient.py`.)

comex:
- Only apply when actually running under py2app, as opposed to whenever
  running on macOS.
- Instead of setting PATH to a fully hardcoded value, just append
  /usr/local/bin if not already present.  That way any non-default
  choices by the user are respected.
- Unset PYTHONPATH instead of setting it based on /usr/bin/python's
  sys.path.  mpv or other players could end up invoking any copy of
  Python, so there's no sensible value to set it to, but it's (normally)
  fine to leave it blank.  The important thing is that we don't keep the
  value set by py2app, pointing to our own Contents/Resources directory.
  (It should be safe to change PYTHONPATH at this point without
  affecting Syncplay's own execution, since Python only checks it at
  startup.)
- Unset other Python environment variables instead of just PYTHONPATH,
  because py2app sets a number of them (including
  PYTHONDONTWRITEBYTECODE and PYTHONHOME).
- Do the environment variable patchup when starting Syncplay, so it
  applies to all players rather than just mplayer-based ones.

Fixes #276.

### DIFF
--- a/syncplay/client.py
+++ b/syncplay/client.py
@@ -40,8 +40,9 @@ from syncplay.constants import PRIVACY_SENDHASHED_MODE, PRIVACY_DONTSEND_MODE, \
     PRIVACY_HIDDENFILENAME
 from syncplay.messages import getMissingStrings, getMessage
 from syncplay.protocols import SyncClientProtocol
-from syncplay.utils import isMacOS
+from syncplay.utils import isMacOS, fixMacEnvironmentVars
 
+fixMacEnvironmentVars()
 
 class SyncClientFactory(ClientFactory):
     def __init__(self, client, retry=constants.RECONNECT_RETRIES):
--- a/syncplay/players/mplayer.py
+++ b/syncplay/players/mplayer.py
@@ -339,20 +339,7 @@ class MplayerPlayer(BasePlayer):
             env = os.environ.copy()
             if 'TERM' in env:
                 del env['TERM']
-            # On macOS, youtube-dl requires system python to run. Set the environment
-            # to allow that version of python to be executed in the mpv subprocess.
-            if isMacOS():
-                try:
-                    pythonLibs = subprocess.check_output(['/usr/bin/python', '-E', '-c',
-                                                          'import sys; print(sys.path)'],
-                                                          text=True, env=dict())
-                    pythonLibs = ast.literal_eval(pythonLibs)
-                    pythonPath = ':'.join(pythonLibs[1:])
-                except:
-                    pythonPath = None
-                if pythonPath is not None:
-                    env['PATH'] = '/usr/bin:/usr/local/bin'
-                    env['PYTHONPATH'] = pythonPath
+
             if filePath:
                 self.__process = subprocess.Popen(
                     call, stdout=subprocess.PIPE, stdin=subprocess.PIPE, stderr=subprocess.STDOUT,
--- a/syncplay/utils.py
+++ b/syncplay/utils.py
@@ -506,3 +506,27 @@ class RandomStringGenerator(object):
 
 class NotControlledRoom(Exception):
     pass
+
+def fixMacEnvironmentVars():
+    if getattr(sys, 'frozen', None) != 'macosx_app':
+        return
+    # py2app, used on macOS, sets a number of environment variables
+    # which are meant for our own Python interpreter but will leak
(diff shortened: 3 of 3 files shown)


syncplay/client.py
```diff
@@ -40,8 +40,9 @@
     PRIVACY_HIDDENFILENAME
 from syncplay.messages import getMissingStrings, getMessage
 from syncplay.protocols import SyncClientProtocol
-from syncplay.utils import isMacOS
+from syncplay.utils import isMacOS, fixMacEnvironmentVars
 
+fixMacEnvironmentVars()
```
albertosottile: I would not modify the os copy of the environment, given that other parts of the code interact with it. Perhaps you can make the method return a modified copy of the environment and use that for the mpv subprocess.

syncplay/players/mplayer.py
```diff
@@ -339,20 +339,7 @@ def __init__(self, playerController, playerPath, filePath, args):
             env = os.environ.copy()
```
albertosottile: As stated before, I would restructure the method to modify this copy (or return a modified copy) of the environment.

Et0h:
@comex Have you considered the comments by @albertosottile? As far as I can see these concerns need to be addressed and we would need to do some teesting before this PR is ready to be merged.

albertosottile:
Closing this as we did not receive any additional information since May. Please feel free to reopen it anytime.

PR #279, opened 2020-03-22
Remove ruby reinstall workaround for openssl, as not needed

palfrey:
So it turns out the workaround hack in #275 isn't needed any more, probably due to the homebrew changes there restricting the alterations to the filesystem. https://travis-ci.org/github/palfrey/syncplay/jobs/665559940 demos a upload being done, and other than it failing due to authorisation issues (which hopefully are just due to it being on my fork, not the main repo) it should hopefully work now!

### DIFF
--- a/travis/macos-deploy.sh
+++ b/travis/macos-deploy.sh
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+set -ex
+
 mkdir dist/Syncplay.app/Contents/Resources/English.lproj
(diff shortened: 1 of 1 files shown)


albertosottile:
Just saw this MR, I was about to try this myself. Let's see if that is the right call. Thanks for the fast fix!

palfrey:
> Just saw this MR, I was about to try this myself. Let's see if that is the right call. Thanks for the fast fix!

Looks like the build passed, so LGTM. Thanks for the tool! It's been useful to me at the past, and so I figured I'd try and fix the build.

PR #291, opened 2020-04-12
Add Brazilian Portuguese translation

AtilioA:
Add new 'pt_BR' language to 'language-argument'


Et0h:
Wow, that's great. Thanks for all your work on this! :D

Issue #261, opened 2019-11-17
upcoming mpv release is incompatible due to API change


wiiaboo:
https://github.com/mpv-player/mpv/commit/f57f13ceb031e5a11a4a2de226ee1f9b211ddb64

No date for the removal of `input-file`, but it's finally being deprecated.

Et0h:
Thanks for letting us know @wiiaboo

I have had an initial look at https://github.com/mpv-player/mpv/blob/master/DOCS/man/ipc.rst and it looks like moving to input-ipc-server would require moving to JSON and might also involve having to work around some UTF-8 quirks and might have some additional issues working on Windows, so not a straightforward switch from one command line switch to another.

Ideally we would want any change to be done in a way which maintained backwards compatibility with mpv >= 0.6.2 (which as per https://syncplay.pl/guide/install/ is the earliest version of mpv we say we support).

wiiaboo:
I don't think it's sane to still support that old an mpv. At that point, still supporting mplayer/mplayer2 makes as much sense. Might as well just create a new mpv thing and that should be supported since at least the moment named pipes were supported in Windows (~[v0.17.0](https://github.com/mpv-player/mpv/commit/5bf473d5ca3b1a8f0902d08799b28512eb81070b)).

I've been using a python script to send URLs from the browser to a single-instance mpv for years now, so named pipes should be fine. Not sure about two-way communication though. My script is one-way only:

```python
#!python3
#!/usr/bin/env python3
#mpv-append.pyw
import sys, subprocess, shlex, codecs, os
from time import strftime
import argparse

log_path = os.path.join(os.path.dirname(os.path.realpath(__file__)), 'umpv.log')

parser = argparse.ArgumentParser(prefix_chars='-/')
parser.add_argument('mpvopts',
                    help='opts passed to mpv if already open',
                    nargs='+')
parsed_args = parser.parse_args(sys.argv[1:])

with codecs.open(log_path, 'a', 'utf-8') as log_file:
    if len(sys.argv[1:]) > len(parsed_args.mpvopts):
        log_file.write('[{}] {} ⇒ "{}"\n'.format(strftime("%Y-%m-%d %H:%M:%S"),
            sys.argv[1:], '" "'.join(parsed_args.mpvopts)))
    else:
        log_file.write('[{}] "{}"\n'.format(strftime("%Y-%m-%d %H:%M:%S"),
            sys.argv[1:], '" "'.join(parsed_args.mpvopts)))

try:
[...]

PR #296, opened 2020-04-14
 Use mpv's new create_osd_overlay API

FichteFoll:
The new API was added in https://github.com/mpv-player/mpv/commit/07287262513c0d1ea46b7beaf100e73f2008295f for mpv 0.31 and doesn't add new functionality, but it makes the existing private interface publicly available, so it should be used eventually. In my journey of fixing the performance regression, I also ported syncplay's code to use this new API.

This was rebased on top of #295 in order to merge cleanly.

I did not change any code regarding the minimum version requirement of mpv because I didn't know how you would like to do that. I see that there is still code supporting some old mpv client with a version lower than 0.6. Please decide for yourself how you want to handle this or whether you want to accept it at all. There is a legacy layer that will map to the new API without any functional differences (at least as of now).

### DIFF
--- a/syncplay/resources/syncplayintf.lua
+++ b/syncplay/resources/syncplayintf.lua
@@ -50,7 +50,6 @@ local FONT_SIZE_MULTIPLIER = 2
 
 local chat_log = {}
 
-local assdraw = require "mp.assdraw"
 
 local opt = require 'mp.options'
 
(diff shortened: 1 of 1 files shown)


daniel-123:
#261 is already going to have to bump minimum supported version to at least 0.17.

That said the policy we'd like to keep is to support player versions at least as old as found in latest LTS releases of popular distros. Which in case of mpv right now means 0.27 - so before this is merged it needs to cleanly handle older version.

Initial workaround on the other hand probably just needs a quick review.

FichteFoll:
Since there is a compatibility layer on the side of mpv, I would rather not want to duplicate this even into scripts unless necessary (I. e. removed by mpv) and reject it.

daniel-123:
I agree. Though instead of rejecting this outright I think it makes sense to simply keep it around until it's needed (mpv removes/changes old API) or until the point where raising minimum supported mpv version to 0.31 won't be a problem.

FichteFoll:
Sure, we can leave it floating around here. I have my forked repositories for pull requests in a separate org anyway for easier organization.

PR #306, opened 2020-05-02
client: add simple wildcard matching mechanism for trusted domains

### LINKED ISSUE #305: Wildcards in trusted domains
eg. ``*.bandcamp.com``
Some websites are kinda annoying when it comes to this. I don't think the security trade-off is that important if it has to be manually entered?

seitokaichou:
closes #305

### DIFF
--- a/syncplay/client.py
+++ b/syncplay/client.py
@@ -506,6 +506,12 @@ class SyncplayClient(object):
         return False
 
     def isURITrusted(self, URIToTest):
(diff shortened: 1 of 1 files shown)


Et0h:
Cheers. Tested it and it seems to allow you to add *.xyz.com as trusted domain and it will then accept abc.xyz.com as a trusted domain, thus resolving the use cases provided by #305

PR #310, opened 2020-05-11
Move to JSON IPC API for mpv using iwaltons3's library (#261)

Et0h:
This should allow Syncplay to support mpv >0.30.0. Raises minimum mpv version to >= 0.17.0.

Many thanks to @iwalton3 for his Python MPV JSONIPC library which is available from https://github.com/iwalton3/python-mpv-jsonipc/

### DIFF
--- a/syncplay/constants.py
+++ b/syncplay/constants.py
@@ -240,7 +240,6 @@ MPV_ARGS = {'force-window': 'yes',
             'hr-seek': 'always',
             'keep-open': 'yes',
             'input-terminal': 'no',
-            'terminal': 'yes',
             'term-playing-msg': '<SyncplayUpdateFile>\nANS_filename=${filename}\nANS_length=${=duration:${=length:0}}\nANS_path=${path}\n</SyncplayUpdateFile>',
             }
 
--- a/syncplay/constants.py
+++ b/syncplay/constants.py
@@ -235,9 +235,16 @@ USERLIST_GUI_USERNAME_COLUMN = 0
 USERLIST_GUI_FILENAME_COLUMN = 3
 
 MPLAYER_SLAVE_ARGS = ['-slave', '--hr-seek=always', '-nomsgcolor', '-msglevel', 'all=1:global=4:cplayer=4', '-af-add', 'scaletempo']
-MPV_ARGS = ['--force-window', '--idle', '--hr-seek=always', '--keep-open']
-MPV_SLAVE_ARGS = ['--msg-level=all=error,cplayer=info,term-msg=info', '--input-terminal=no']
-MPV_SLAVE_ARGS_NEW = ['--term-playing-msg=<SyncplayUpdateFile>\nANS_filename=${filename}\nANS_length=${=duration:${=length:0}}\nANS_path=${path}\n</SyncplayUpdateFile>', '--terminal=yes']
+MPV_ARGS = {'force-window': 'yes',
+            'idle': 'yes',
+            'hr-seek': 'always',
+            'keep-open': 'yes',
+            'input-terminal': 'no',
+            'terminal': 'yes',
+            'term-playing-msg': '<SyncplayUpdateFile>\nANS_filename=${filename}\nANS_length=${=duration:${=length:0}}\nANS_path=${path}\n</SyncplayUpdateFile>',
+            }
+
+
 MPV_NEW_VERSION = False
 MPV_OSC_VISIBILITY_CHANGE_VERSION = False
 MPV_INPUT_PROMPT_START_CHARACTER = "〉"
--- a/syncplay/constants.py
+++ b/syncplay/constants.py
@@ -236,7 +236,7 @@ USERLIST_GUI_FILENAME_COLUMN = 3
 
 MPLAYER_SLAVE_ARGS = ['-slave', '--hr-seek=always', '-nomsgcolor', '-msglevel', 'all=1:global=4:cplayer=4', '-af-add', 'scaletempo']
 MPV_ARGS = ['--force-window', '--idle', '--hr-seek=always', '--keep-open']
-MPV_SLAVE_ARGS = ['--msg-level=all=error,cplayer=info,term-msg=info', '--input-terminal=no', '--input-file=/dev/stdin']
+MPV_SLAVE_ARGS = ['--msg-level=all=error,cplayer=info,term-msg=info', '--input-terminal=no']
 MPV_SLAVE_ARGS_NEW = ['--term-playing-msg=<SyncplayUpdateFile>\nANS_filename=${filename}\nANS_length=${=duration:${=length:0}}\nANS_path=${path}\n</SyncplayUpdateFile>', '--terminal=yes']
 MPV_NEW_VERSION = False
 MPV_OSC_VISIBILITY_CHANGE_VERSION = False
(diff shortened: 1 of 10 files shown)


daniel-123:
I have tested it with mpv 0.23, 0.29.0 and 0.32 - it works except for needing to wait for ipc

iwalton3:
The library currently assumes MPV is running and has IPC available if `start_mpv` is set to false. (That is intended for connecting to a copy of MPV the user already has running, like GNOME MPV.) The managed mpv allows you to pass command arguments (when you initialize the MPV constructor), will generate a randomly named Linux socket/Windows pipe (if not specified), and will wait for IPC to open.

Of course you can verify that the socket/pipe is already open before calling the constructor. Just be careful as Windows names pipes behave oddly if you check them twice in a short period of time.

Et0h:
I've tried moving over to managed mpv as per @iwalton3's suggestion to see if that fixes the issue raised by @daniel-123.

When testing the latest commit please pay close attention to:
* Errors during initialisation (e.g. due to Syncplay trying to access things too soon)
* Handling of additional command line arguments specified by Syncplay users (as I moved from a list to a dictionary)
* Handling of opening mpv with a filename specified in Syncplay (as I moved over to the delayedFilePath system for all operating systems)
* Handling of Syncplay detecting and reacting to mpv being closed (as I had to change the relevant code)

iwalton3:
If there is a problem with it detecting events such as window close, you can catch them with an event handler: https://github.com/iwalton3/jellyfin-mpv-shim/blob/master/jellyfin_mpv_shim/player.py#L111

Et0h:
@iwalton3 In terms of event handling, if two scripts both register for the same on_key_press event then it will only actually honour one of them? That was my vague recollection from looking into it a long time ago, but maybe I'm wrong or out of date.

iwalton3:
Yes if you bind a key, it will prevent all behavior from all previous key binds, including the default ones.

syncplay/players/python_mpv_jsonipc/python_mpv_jsonipc.py
```diff
@@ -0,0 +1,607 @@
...
+    def __init__(self, ipc_socket, callback=None):
+        """Create the wrapper.
+
+        *ipc_socket* is the path to the socket.
+        *callback(json_data)* is the function for recieving events.
+        """
+        self.ipc_socket = ipc_socket
+        self.callback = callback
+        self.socket = socket.socket(socket.AF_UNIX)
+        self.socket.connect(self.ipc_socket)
```
daniel-123, line 123: Testing whether ipc socket is ready is needed before this line.
On my own system it takes an extra delay of 20-30 ms inserted here.

syncplay/constants.py
```diff
@@ -235,9 +235,16 @@ def getValueForOS(constantDict):
...
 MPLAYER_SLAVE_ARGS = ['-slave', '--hr-seek=always', '-nomsgcolor', '-msglevel', 'all=1:global=4:cplayer=4', '-af-add', 'scaletempo']
-MPV_ARGS = ['--force-window', '--idle', '--hr-seek=always', '--keep-open']
-MPV_SLAVE_ARGS = ['--msg-level=all=error,cplayer=info,term-msg=info', '--input-terminal=no']
-MPV_SLAVE_ARGS_NEW = ['--term-playing-msg=<SyncplayUpdateFile>\nANS_filename=${filename}\nANS_length=${=duration:${=length:0}}\nANS_path=${path}\n</SyncplayUpdateFile>', '--terminal=yes']
+MPV_ARGS = {'force-window': 'yes',
+            'idle': 'yes',
+            'hr-seek': 'always',
+            'keep-open': 'yes',
+            'input-terminal': 'no',
+            'terminal': 'yes',
```
daniel-123, line 243: `'terminal': 'yes'` causes mpv stdout to be printed into stdout of Syncplay. It includes stuff that goes through ipc.

It might be useful for debugging, but I don't think it should be the default. Removing this option doesn't seem to influence how Syncplay works.

iwalton3: It looks like Syncplay is using the debug handler that gets the events, so there should be no reason to have that turned on. It might be left over from the former system, where I think it was capturing STDOUT as part of the communications.

daniel-123: Yes, that's the case. As it stands now though it's going to severely interfere with --no-gui mode of Syncplay.

It does seem that this change prevents Syncplay from realizing immediately that mpv process quit. There is a 60 second timeout so it eventually catches up, but it's not what one would expect.

So ultimately either this terminal output should be suppressed or detection of when mpv dies should be changed.

iwalton3: I currently use the library in my projects with the assumption mpv would run once for the life of the program, and then you simply send it play and stop commands. (By default the library starts with no GUI.)

My library should be able to handle mpv being restarted if terminated by syncplay, but I don't remember the other python mpv library working reliably when I restarted MPV, hence that design.

iwalton3:
> Quitting mpv raises an exception
> `EventHandler caught exception from (<bound method MpvPlayer.mpv_log_handler of <syncplay.players.mpv.MpvPlayer object at 0x7f197c031d30>>, ('info', 'cplayer', 'Exiting... (Quit)')).`
> 
> Expected behavior would be to just quit Syncplay instead.

The best way to handle this would probably be to rebind `q` and `CLOSE_WIN` to trigger syncplay to quit.

daniel-123:
> > Quitting mpv raises an exception
> > `EventHandler caught exception from (<bound method MpvPlayer.mpv_log_handler of <syncplay.players.mpv.MpvPlayer object at 0x7f197c031d30>>, ('info', 'cplayer', 'Exiting... (Quit)')).`
> > Expected behavior would be to just quit Syncplay instead.
> 
> The best way to handle this would probably be to rebind `q` and `CLOSE_WIN` to trigger syncplay to quit.

I feel iffy about rebinding keys. It won't do anything if somebody uses different key to quit mpv (or just clicks the X). It also won't handle the situation where mpv crashes for some reason.

I think it's just much simpler to ensure consistent behavior if Syncplay quits the instant mpv process is no longer there regardless of the reason.

iwalton3:
> Edit: an exception further down is a simple BrokenPipeError on the ipc socket, which should be reasonably foolproof way of detecting that mpv is gone.

Yeah that is probably your best bet, since it still works even if it is connecting to a copy the user launches. That would have to be caught in the separate pipe implementations and sent back as some kind of event (since they're separate threads).

iwalton3:
It looks like the actual "BrokenPipeError: [Errno 32] Broken pipe" doesn't happen until you try call something on the MPV instance after it has crashed. (Currently the broken pipe is ignored in the actual socket implementations.) I'll add a callback you can register to know when MPV terminates.

iwalton3:
Please see: https://github.com/iwalton3/python-mpv-jsonipc/releases/tag/v1.1.11

blaenk:
Thanks so much for working on this @Et0h! Can we expect any kind of release (stable or otherwise) once this is merged? I would love to try it out at that point.

Et0h:
> Thanks so much for working on this @Et0h! Can we expect any kind of release (stable or otherwise) once this is merged? I would love to try it out at that point.

If no showstoppers are reported then I expect this PR will be merged into the main branch before the end of this week after it has been tested on Mac. I'm hoping to get at a release candidate of Syncplay v1.6.5 out before the end of the month.

PR #312, opened 2020-05-16
client: actually fix the wildcard matching

seitokaichou:
This is more robust.
The old code had a mistake as well, which is very embarrassing.


Et0h:
All's well that ends well.

Issue #262, opened 2019-11-30
Feature request: Queue videos (and more controls?) without graphical interface


Slider-Whistle:
I imagine this would work through a command called "q" or "queue". Could assume local file if URI is "file:" or none, passed to media player as-is if it's HTTP or FTP or whatever.
I can't see much else crucial functionality that's missing without the GUI, but I guess it might be a good goal to make anything possible through the chat interface.

Et0h:
Are you talking about adding files to the playlist through the console interface? What is the use case for this?

Slider-Whistle:
Maybe through the console interface if the user has an active terminal, but I mean mostly through the chat itself. I just think that people would find it more convenient.

Et0h:
Is your idea that queue would add a filename/URL to the bottom of the playlist? That could make it slightly quicker to add YouTube videos, etc, to the playlist.

Slider-Whistle:
Yeah, as long as the user has permissions. I don't think such a command exists in Cytube or similar services, but it's common enough with chatbots and stuff that I feel like I've gotten used to it.

FichteFoll:
Related to #272 and #232, where I want to add videos to the playlist from the clipboard and from an external program through CLI respectively.

If a command interface was implemented for the chat and that would also be available from the CLI to execute one-off commands, that would be sweet.

One thing to consider would be adding multiple URLs or files, separated by a newline. The chat is inherently based on single line messages and not suited for adding multiple URLs, unless it became multi-line aware, supporting auto-resizing and a shift+enter binding to insert a newline instead of sending the message, for example.

TWINGSISTER:
Thanks for this software, it is great!
This could be of unvaluable interest in classrooms.
Whenever you have a ICT labo with some PC with loudspeakers and want to show a video you need student to use headphones or there will be a noisy mess.
Synching all the PC will solve the problem and help the teacher to stop playing to add some comment.

It would be nice if the server can expose an http API with three or four basic commands like
-Play this URL starting at this time (may be either video| mp3 audio )
-Stop/Fade out Playing now|at time
-Resume Playing
-Set volume to...

Then this, imho, will be the right start to a very big deal!

If you can provide Python code for this i think I can translate this in PHP and make this a Wordpress plugin and
you will have the support for internet radio with unlimited possibilities and no copyright infringment. Wordpress will just store and broadcast some (possibly hashed) youtu.be URLs  (that is something that a lot of WP plugins are already doing with no encryption) 
[...]

PR #316, opened 2020-05-25
add videos to playlist from chat

csandras05:
(Partially) solves issue #262 - adding videos to playlist from terminal (without gui) is still not possible.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,10 @@ jobs:
       dist: xenial
       os: linux
       env: BUILD_DESTINATION=snapcraft
+    - language: minimal
+      dist: focal
+      os: linux
+      env: BUILD_DESTINATION=deb
     - language: python
       os: linux
       dist: xenial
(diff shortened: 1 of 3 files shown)


Et0h:
Thanks for the PR.

It might be better if users could add any file to the playlist (as users can when you edit the playlist), which could be useful if they are still downloading the file they were hoping to play.

I've not tested it much, but I think this can be done by making the following changes:
- Remove the various checks in `elif command.group('command') in constants.COMMANDS_QUEUE:` (within _consoleUI.py_) and simply using  `self._syncplayClient.ui.addFileToPlaylist(filename)`.
- In `def addFileToPlaylist(self, filePath, index=-1):` in _gui.py_, replace` if os.path.isfile(filePath):` with `if not isURL(filePath):` and replace `elif isURL(filePath):` with `else:`

Note: The check in `elif command.group('command') in constants.COMMANDS_QUEUE:` for if the parameter is blank could potentially be retained, but the other commands do not have such checks so perhaps you should either skip this error or introduce a generic error for a lack of parameter in the other languages for all of the commands for consistency.

Some language issues:
- `"No file/url given"` - All messages should be in messages_en.py with stubs in other languages (with #TODO: Translate). URL should be capitalised. That said, as above, this error might not want to be retained.
- `"\tq [file/url] - add file or url to bottom of playlist"`, - Should have stubs in other languages. URL should be capitalised.

PR #317, opened 2020-05-31
I corrected some not  transltated on pt-BR and added pt-PT

artur309:
I corrected some not translated on pt-BR.

I added the pt-PT language since it changed the way the phrases and some words pt-BR does.


Et0h:
Thank you for your work on this. It isn't ready to be merged yet because if you look at https://ci.appveyor.com/project/Et0h/syncplay-88efr/builds/33227583 then you will see that the compilation of the NSIS script failed.

I am not 100% sure why, but I checked my NSIS folder and it does not contain `PortuguesePT.nlf` but it does contain `Portuguese.nlf` so the issue might be the filename cited needs changing.

It might also be the case that LANG_PORTUGUESEPT needs to just be LANG_PORTUGUESE, but I'm not 100% sure.

artur309:
> Thank you for your work on this. It isn't ready to be merged yet because if you look at https://ci.appveyor.com/project/Et0h/syncplay-88efr/builds/33227583 then you will see that the compilation of the NSIS script failed.
> 
> I am not 100% sure why, but I checked my NSIS folder and it does not contain `PortuguesePT.nlf` but it does contain `Portuguese.nlf` so the issue might be the filename cited needs changing.
> 
> It might also be the case that LANG_PORTUGUESEPT needs to just be LANG_PORTUGUESE, but I'm not 100% sure.

Alright I will try see to

Et0h:
Just checked and it now gives the error `VIAddVersionKey: "/LANG=${LANG_PORTUGUESEPT}" is not a valid language code!`

To solve this you need to update `buildPy2exe.exe` to replace all references of `${LANG_PORTUGUESEPT}` to `${LANG_PORTUGUESE}`

artur309:
> Just checked and it now gives the error `VIAddVersionKey: "/LANG=${LANG_PORTUGUESEPT}" is not a valid language code!`
> 
> To solve this you need to update `buildPy2exe.exe` to replace all references of `${LANG_PORTUGUESEPT}` to `${LANG_PORTUGUESE}`

Ok, thanks for bug reporting. I will try do that.

artur309:
It's done it have passed all checks :)

syncplay/messages_pt_BR.py
```diff
@@ -126,8 +126,8 @@
     "invalid-port-config-error": "A porta deve ser válida",
     "empty-value-config-error": "{} não pode ser vazio",  # Config option
 
-    "not-json-error": "Não é uma string codificada como json\n",
-    "hello-arguments-error": "Not enough Hello arguments\n",  # DO NOT TRANSLATE
+    "not-json-error": "Não é uma string codificada como JSON\n",
+    "hello-arguments-error": "Falta de argumentos Hello\n",  # DO NOT TRANSLATE
```
AtilioA, line 130: I don't think this line should be translated...

Et0h: Reverted 👍

AtilioA:
@Et0h Maybe `hello-arguments-error` shouldn't be included in the translation files altogether?

Et0h:
> @Et0h Maybe `hello-arguments-error` shouldn't be included in the translation files altogether?

Could do, although then I'd need to add extra code to avoid a build warning for a lack of a translation message.

PR #319, opened 2020-06-03
add files to playlist from chat

csandras05:
Had some issues with git, reopening PR #316 with the requested changes.

### DIFF
--- a/syncplay/constants.py
+++ b/syncplay/constants.py
@@ -124,6 +124,7 @@ COMMANDS_HELP = ['help', 'h', '?', '/?', r'\?']
 COMMANDS_CREATE = ['c', 'create']
 COMMANDS_AUTH = ['a', 'auth']
 COMMANDS_TOGGLE = ['t', 'toggle']
+COMMANDS_QUEUE = ['queue', 'q', 'add']
 MPC_MIN_VER = "1.6.4"
 MPC_BE_MIN_VER = "1.5.2.3123"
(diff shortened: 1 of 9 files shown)


daniel-123:
Trying to call the "q" command to add file to playlist while in CLI mode (`--no-gui`) results in `TypeError: addFileToPlaylist() takes 1 positional argument but 2 were given` in client.py on line 1550.

Et0h:
@daniel-123 I don't think that is due to this PR, I think it is because the `def addFileToPlaylist(self):` stub in `consoleUI.py` needs to accept more parameters (e.g. it should be `def addFileToPlaylist(self, filePath, index=-1):`). We could also make it give an appropriate error message saying that playlists are not currently supported in console made (rather than just doing a `pass`).

daniel-123:
I've tried out the CLI mode for a bit. Given it's current state with effective lack of playlist support I think I'd be happy with a solution that simply doesn't crash the client.

Fleshing out CLI mode to include media directories support and eventually proper playlist support is a possible separate goal, but given that GUI is by far the primary mode of use I would see it as low priority and out of scope for this PR.

Et0h:
#327 uses #319 as its base. At present I'm minded to go with a tweaked version of #327 instead of #319 because as far as I'm aware #327 contains the benefits of #319 and more.

Et0h:
Relevant elements incorporated into #327 which has now been merged.

PR #320, opened 2020-06-04
MPV: Add the possibility to control which socket to use

TaPO4eg3D:
Allow MPV users to set a custom socket via __input-ipc-server__ flag. This will make possible simultaneous use of Syncplay and other tools like SVP-project


TaPO4eg3D:
Yes, I tested it. To make it work with SVP-project it is enough to just specify in the player flags:
--input-ipc-server=/tmp/mpvsocket

Et0h:
I'll need to test this, but it seems like a good improvement.

Et0h:
Test it and it all looks good :)

Issue #336, opened 2020-07-17
Feature request: Room history mechanism


Gbd199:
Provide a dropdown list with the names of previous rooms.
Requires configuration parsing changes/additions and gui logic enhancement.

Gbd199:
Opened a pull request #337

Gbd199:
Reopening for notice and feedback

Gbd199:
@Et0h  Can you please take a look and share your thoughts?

Et0h:
I've posted my thoughts to #337 and the discussion can continue there.

Et0h:
Okay, so I've accepted #337 and tweaked it with 3c7ca9706d7dfb00d7e1a84ebae7a534d54d42b9

Some things to note about changed from #337 (and a few requests for feedback):
* For now I've made it so that you have to manually add rooms to the room list. There could be an optional in the future to allow for rooms to be automatically added, but it would then add complexity to the GUI so it might be fine just leaving it as manual.
* I've renamed it 'room list' rather than 'room history' as it is a user-editable list, not just a history.
* I've added the combo box to the main GUI and added the option to change the room list to the 'window' portion of the menu bar. I didn't add it to the main GUI because I didn't want to add too many options there (as it). However, I could add a "Show edit room button" option to the Window menu (off by default) if people wanted, in line with what I do for playback controls and auto-play buttons. Thoughts/opinions?
* I've moved the edit button in the config window to the right to be more in line with normal UI conventions (even if it is less space efficient).
* For now I've not changed it to automatically change the default room to the last one you joined as I'm not sure if that's the behaviour most people want. Thoughts/opinions?

Et0h:
Changes:
* Room names are now saved by default, this is changeable in the "Add rooms you join to the room" list option in "Misc". I was thinking that it could be the same option for if we have an option in the future to save server addresses.
* When you create a managed room, it also saved the version of the room name which includes the room password.
* Room names are now listed alphabetically. Previous behaviour was to have the most recently joined rooms at the top.

Happy for feedback on both of these changes.

Et0h:
As this feature has now been introduced in Syncplay 1.6.7 I will close this issue. Thanks to everyone who has contributed!

Gbd199:
Very welcomed changes and decisions. 
Thanks a lot for advancing the issue @Et0h! Awesome to see the mechanism in an official version

PR #337, opened 2020-07-17
Room history

Gbd199:


Et0h:
Hi,

I have long wanted for a feature such as this so thanks for helping to move it forwards.

Firstly, I've tested and it works. The behaviour is that it lists the rooms you connect to the server with and then provides them as options from the GuiConfig window for future connections, irrespective of server. As far as it goes, this is a reasonable behaviour because trying to make it per-server arguably adds unnecessary complexity when people might use the same room names across multiple servers

Suggestions for consideration/discussion:
* The ability to delete a room from the list of rooms.
* Allow users to select from the list of rooms once connected, e.g. by making the textbox in the main Syncplay window next to 'Join room' also be a combo box.
* Remember rooms you connect to from the main Syncplay window
* When you create a managed room, also remember the version with the auto-identification in it, e.g. +RoomName:868A81271506:NX-914-891
* An alternative approach to auto-remembering the room names would be to have a 'favourites' system where, once connected, you can add a room to your list of favourites and then manage that list. This will prevent the list from getting clogged up so requires less user action to clear it, but on the other hand does require users to manually click to remember a room which they might forget to do. Which is more convenient and intuitive?

Gbd199:
thanks for the feedback, I'll see what I can do over the weekend

Gbd199:
I decided to go with a dialog for the rooms. This gives the user complete and simple control over the rooms and did not require too many changes. Due to the simple control I think adding a 'favorite rooms' mechanism is pointless.

Et0h:
Allowing people to edit, add and delete rooms via a multiline textbox dialog makes sense, not least because it is in line with what we do for trusted domains and the playlist. I'll try out your latest version and give it some consideration. Something I'll be thinking about is whether there needs to be an easy way to enable/disable the feature automatically remembering room names.

Et0h:
Thought about it, and I think the only thing it really needs is to be able to turn off auto-remembering. If nobody has any further comments/objections then when I have time I'll merge the changes and add any final tweaks myself.

Et0h:
Further discussion take take place on #336

PR #348, opened 2020-08-13
Changed mpvnet icon from .ico to .png for consistency

mentofa123:
I made a small change for some consistency. This changes the mpvnet icon from a .ico file to a .png file like the rest of the other icons.


mentofa123:
This fixes the bug where the icon of mpvnet doesn't show up in the settings. 
As a reference:
- Before 
image

- After 
image

Et0h:
I'll need to test this, but it seems like a good improvement.

Issue #361, opened 2020-10-21
Options on start up


MarisKori:
I chose "do not show this dialog on start up" (about options), but it is shown again and again each time I start the Syncplay.
v1.6.6 beta 2 (release 89)

Et0h:
Do you mean the _"Don’t always show the Syncplay configuration window"_ option?

As noted in the tooltip text and the guide at https://syncplay.pl/guide/client/ this option only means that Syncplay won't open the configuration window if you open a file with Syncplay unless you need to specify configuration options. If you are running Syncplay directly then it will always display the configuration window irrespective of what you choose for that option.

MarisKori:
Yes, I meant exactly than option. I translated it from Russian. And in Russian it sounds like "Don't show the Syncplay configuration window at all".

Et0h:
I just checked and you're right, according to Google "Не показывать больше этот диалог" translates to "Don't show this dialog again". Can you please review this @and-semakin ?

Et0h:
@MarisKori If you let me know what the Russian translation should be then I can change it. @and-semakin can then tweak if it needed.

PR #363, opened 2020-10-31
Update Russian translation

OlegSmelov:
Translate some of the messages into Russian.


Et0h:
Thanks for your work on this @OlegSmelov. Could you also please also update the translation for forceguiprompt-label in line with the discussion in #361

PR #367, opened 2020-11-25
Create icons folder if it doesn't exist

Lctrs:
If it doesn't exist first, `hicolor` will be stripped from the path by the `cp` command

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -38,6 +38,7 @@ common:
 	-mkdir -p $(LIB_PATH)/syncplay/syncplay/resources/lua/intf
 	-mkdir -p $(APP_SHORTCUT_PATH)
 	-mkdir -p $(SHARE_PATH)/pixmaps/
+	-mkdir -p $(SHARE_PATH)/icons/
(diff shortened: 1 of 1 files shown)


Et0h:
I'm expecting this is a non-controversial change, but as I use Windows it would be useful if someone with makefile experience can confirm that this is the correct approach.

Et0h:
@Lctrs What OS are you using which you believe exhibits this behaviour (of cp not creating the hicolor folder when the icon folder does not exist) and have you confirmed it?

Lctrs:
When trying to create a flatpak for the client. The icons folder doesn't exist in this case beforehand, and the `cp` command rename the `hicolor` folder to `icons`, resulting in an invalid tree.

daniel-123 (approved):
I've not seen this crop up up until now mostly because the assumption behind this installer script was that it will typically run on a "normal" operating system. Which pretty much always already has some software with icons, so I didn't think about situation where it doesn't exist.

It think that overall this is completely reasonable change. Though we don't have an official flatpak build process, so there is some risk of your package breaking unintentionally with some other changes in future.

PR #369, opened 2020-11-29
Move python_mpv_jsonipc in our vendor folder

albertosottile:
We embed a copy of`python_mpv_jsonipc` (https://github.com/iwalton3/python-mpv-jsonipc) in our package to control mpv and IINA. Since this is external code, it would be better to relocate it in the `vendor` folder that we already have in out package, for this purpose.


Et0h (approved):
Happy with this change. Tested it, and it works fine for me on Windows (including when I build an installer and run Syncplay from that).

PR #372, opened 2020-12-19
Update Russian translation

OlegSmelov:
I've translated all the missing messages into Russian and edited some of the old ones. Some further editing might be necessary to make the tone and terminology consistent.

Continuation of https://github.com/Syncplay/syncplay/pull/363.
Fixes https://github.com/Syncplay/syncplay/issues/361.

@Et0h


Et0h:
It's good to see something which addresses #361, and the other changes seem like a nice bonus. I'll need to find time to test it before it is merged, but I had a peek at the file changes and it looks good at first glance.

Issue #379, opened 2021-01-18
youtube-dl not working when running syncplay with iina or mpv


cordelac:
URL's aren't working when running syncplay with iina or mpv.

> Failed to recognize file format.

However, URL's on both players work fine when outside syncplay. Tested with random youtube links.

OS - Big Sur 11.1 on M1
Syncplay: 1.6.7
youtube-dl: 2021.1.16
mpv: 0.33.0 // iina: 1.1.2 build 128

Et0h:
Not being a Mac users I can't provide much support. I know on some platforms an issue can (or used to) occur if you open a file with Syncplay because that changes the PATH (base location) and so it looks for YTDL in the wrong place.

How are you running Syncplay?

Et0h:
As a workaround you might be able to put something along the files of the following in mpv.conf (with the path being the relevant path to youtube-dl):
`script-opts=ytdl_hook-ytdl_path="/insert/path/here"`

If that works then let me know what worked.

cordelac:
Tried adding the path in mpv.conf, didn't work unfortunately.

I run Syncplay by opening it from Applications? I installed it via Homebrew (though it's the same downloading the dmg).

albertosottile:
My gut says we are witnessing https://github.com/Syncplay/syncplay/commit/c07206c18992c1dca401b30a01b9f0fe54a71df5 on steroids, a.k.a. we are mixing two Python interpreters compiled for two architectures (x86_64 in the Syncplay app and aarch64 in the system provided one used by youtube-dl).

How did you install youtube-dl? Do you use the version of Homebrew for x86_64 or for aarch64? Additionally, if you could open Terminal, run this command:

```

```
and report the result here, you could definitely help us.

(what does that do, you might ask? It finds youtube-dl in your system, asks which interpreter it uses to run, asks the same interpreter which architecture was compiled for.

cordelac:
Oh, I get it. I run Homebrew natively, so in terminal for youtube-dl I got:

`macOS-11.1-arm64-arm-64bit`

I guess I'll need an x86_64 version for Syncplay? In case it's relevant, my mpv is arm, while iina is still Intel

albertosottile:
You would actually need an arm64 version of Syncplay, but... there is none 😦 (see #368)

I guess we now need to think about how we can try to solve/mitigate this on our side. Thanks for reporting this and for your help.

albertosottile:
[...]

PR #381, opened 2021-01-21
Re-enable youtube-dl on macOS plus mpv and IINA

albertosottile:
This is a follow-up to #279. In principle, this issue should have been solved in https://github.com/Syncplay/syncplay/commit/c07206c18992c1dca401b30a01b9f0fe54a71df5 and #360. However, a mixture of changes in the way Homebrew installs `youtube-dl`, plus a mistake in which the environment was (not) transferred to the relevant subprocess, still caused it to show up in 1.6.7.

This PR re-enables the environment patching on macOS and, additionally, slightly modifies how the environment is patched to make this functional also on the arm64 architecture. Credits to @cordelac for having tested this on their machine.

Please test if this affects running mpv on Windows and Linux in any way before merging.

Closes: #379

### DIFF
--- a/syncplay/players/mpv.py
+++ b/syncplay/players/mpv.py
@@ -590,7 +590,7 @@ class MpvPlayer(BasePlayer):
             # to allow that version of python to be executed in the mpv subprocess.
             if isMacOS():
                 try:
-                    env['PATH'] = '/usr/local/bin:/usr/bin'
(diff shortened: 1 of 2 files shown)


Et0h (approved):
I tested this on Windows and could use youtube-dl with mpv and mpv.net without issue, so at the very least it doesn't seem to have broken anything on Windows.

daniel-123 (approved):

Issue #375, opened 2020-12-28
Suggestion: Make a server only package


tippl:
I was setting up a server instance for Syncplay on a server. When i attempted to install the .deb package, the hard dependency on vlc/mpv makes the package manager want to install the whole dependency chain for that (X11 server among other things).

This is really not ideal when trying to install just the server, i don't want to install X11 on a headless machine somewhere in cloud. I solved this by repackaging the deb file and removing the vlc/mpv and pyside2 dependency.

My suggestion is, that there should be a server only package so that you don't have to repackage, but still get the benefits of using a package manager.

Et0h:
I believe the discussion at #335 is relevant, especially the comments by @daniel-123. Linux isn't my area, but my understanding is that if someone is willing to implement and maintain support for a server-only package then they are free to make a PR. You would probably want to wait for PR #374 to be resolved before trying to make a PR implementing this feature, though.

daniel-123:
Because you are second person to ask about this, I actually think that there is enough demand for such package to bother me enough to get around and make it.

Though like @Et0h said - we are currently right in the middle of migrating entire CI process over to different provider in #374 so this will have to wait.

Right now I'm thinking of simply making a no-gui version since that's only matter of removing dependencies and won't require any significant modification to the build script: https://github.com/Syncplay/syncplay/blob/master/travis/deb-script.sh

Last tidbit I'd mention is that server code hasn't been really touched for a good while. Only change it had in recent years is adding TLS support and that's it.

Morsine:
I've been looking for such thing as well!

tcyrus:
I made a fork of syncplay which removed all the client side code a while back. Ii isn't neat, and I changed a large chunk of the code to work with my specific use case, but it's [here](https://github.com/weeb-poly/syncplay-server) if you want it.

daniel-123:
@Morsine @tippl I've made a pull request with server package. Could you test the results? You can download the zip with both server and full packages here: https://github.com/Syncplay/syncplay/suites/1888847905/artifacts/36933740

The package is architecture independent, so it will work on Raspberry Pi as is.

tippl:
@daniel-123 Just tested it on my server. Works great, does not require the gui packages, starts and i'm able to connect to it. I see no issues with it.

daniel-123:
Implemented in #382

Issue #335, opened 2020-07-11
Cannot install on Ubuntu 18.04: python3-pyside2.qtwidgets (>= 5.12.0) not installable


technoabyss:
I'm trying to set up a syncplay server on headless 18.04 Ubuntu.

Running `sudo apt install ./syncplay_1.6.5.deb` gives me the following output:

```
Reading package lists... Done
Building dependency tree
Reading state information... Done
Note, selecting 'syncplay' instead of './syncplay_1.6.5.deb'
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:

The following packages have unmet dependencies:
 syncplay : Depends: python3-pyside2.qtwidgets (>= 5.12.0) but it is not installable
E: Unable to correct problems, you have held broken packages.
```

I'm not sure what I can do about this.

technoabyss:
I ought to mention I managed to set up a server using the AppImage with the `--server` option.

daniel-123:
This is expected. Pyside2 package which is required is not available in Ubuntu 18.04. Oldest Ubuntu version which includes Pyside2 and is compatible with Syncplay deb package is 19.10.

Appimage works because it bundles Pyside2 (and other things) within its binary.

technoabyss:
fair enough. maybe this is a stupid question, but would it be possible to distribute the server binaries separately in releases?

kathelynn:
> This is expected. Pyside2 package which is required is not available in Ubuntu 18.04. Oldest Ubuntu version which includes Pyside2 and is compatible with Syncplay deb package is 19.10.
> 
> Appimage works because it bundles Pyside2 (and other things) within its binary.

Source install also works assuming it requires Qtwidgets >=5.11.0 and no issues so far. Is it safe to match the requirements found in the deb file with the source requirements?

daniel-123:
@kathelynn I think Ubuntu 18.04 doesn't have any qtwidgets version in repositories. At least nothing can be found in [official search](https://packages.ubuntu.com/eoan/python3-pyside2.qtwidgets). Development build of deb package on [bintray](https://bintray.com/syncplay/Syncplay/Syncplay/#files/) has also switched the qtwidgets dependency to include no version information as all versions should work with exception of 5.15 development version where pyside has a bug mentioned in #321.

[...]

PR #382, opened 2021-01-23
Add script to build server only deb package

daniel-123:
Addresses the #375 and #335


daniel-123:
The workflow outputs both deb packages inside single zip. Which might be somewhat inconvenient, but doesn't seem like an issue since no matter what it has to be unpacked anyway. This might change once it's possible to create non-zipped artifacts with Actions.

The server deb package currently includes some excess files which are used only by the client, all the icons etc. There is no trivial way to trim it in maintenance-free way so I left it out from this PR. Later on it can be adjusted by modifying the GNUMakefile.

albertosottile (approved):

albertosottile:
LGTM, I am just not sure if the server package will support also TLS, because I do not know if the `python3-twisted` package includes the dependencies of the `[tls]` extra.

daniel-123:
python3-twisted itself has a hard dependency on python3-openssl. I'm pretty sure there isn't any separate package with just tls parts of twisted on Debian and presumably any of its derivatives.

On both environments I tested this in, the requirements specified were sufficient for TLS support.

PR #404, opened 2021-03-20
Add notice for Python in third party collection file

albertosottile:
On Windows and macOS, we are shipping a copy of Python and of the standard library in the executables. As per the `PSF LICENSE AGREEMENT FOR PYTHON`, we should then display their copyright notice and a copy of the license agreement in our software. 

Hence, this PR adds both these texts in our third party notices file.

In addition, this PR changes the notices file format from RTF to plain text.


Et0h (approved):

Et0h:
Tested the portable version and it works fine for me on Windows 10 64-bit.

daniel-123 (approved):

daniel-123 (approved):

PR #416, opened 2021-05-06
Move GUI dependencies to setup.py extra

FallenWarrior2k:
I wanted to install Syncplay with pip to make use of the syncplay-server entry point. However, when I did, I noticed that setup.py pulls in PySide2 as `install_requires` unconditionally, making that avenue of install undesirable when I only want to run the server in that specific environment. Therefore, I have moved PySide2 to a new `gui` extra in setup.py, so Syncplay can be install by pip without pulling it in.

I have updated the AppImage build script to account for this change, but I haven't been able to find if the installation as a package is used anywhere else. While I did verify that the AppImage still works with the updated build script, I was unable to test the other platforms as I don't have machines for those available right now.

### DIFF
--- a/ci/appimage-script.sh
+++ b/ci/appimage-script.sh
@@ -101,7 +101,7 @@ mv pl.syncplay.syncplay.appdata.xml AppDir/usr/share/metainfo/
 cp "$REPO_ROOT"/syncplay/resources/syncplay.desktop ./pl.syncplay.syncplay.desktop
 
 #export CONDA_PACKAGES="Pillow"
-export PIP_REQUIREMENTS="."
+export PIP_REQUIREMENTS=".[gui]"
 export PIP_WORKDIR="$REPO_ROOT"
 export VERSION="$(cat $REPO_ROOT/syncplay/__init__.py | awk '/version/ {gsub("\047", "", $3); print $NF}')"
 export OUTPUT=Syncplay-$VERSION-x86_64.AppImage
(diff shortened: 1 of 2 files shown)


FallenWarrior2k:
True. I can definitely understand wanting to make the GUI installation as frictionless as possible. It's just that, for me, it felt easier and cleaner to do `pip install git+https://github.com/syncplay/syncplay.git@v1.6.7#egg=syncplay`, instead of having to download an archive, unpack it somewhere (meaning it's not tracked by any package manager; I'm not using Debian/Ubuntu atm, so the deb packages weren't an option), and then having to keep that directory around.

I have to admit that I'm not too intimately familiar with Python packaging. With Cargo (first example that came to mind), this could've easily been done with a default feature, but it seems extras in Python do not support such default semantics (cf. pypa/setuptools#1139).

The method described [here](https://github.com/pypa/setuptools/issues/1139#issuecomment-326565450) could have worked, e.g. by specifying an empty extra `server` and adding a condition `extra != 'server'` to the pyside2 dependency. However, as [this comment](https://github.com/pypa/setuptools/issues/1139#issuecomment-352498909) states, that method/workaround no longer works.

With things as they are, I am not aware of a method to enable this, other than splitting the GUI into a distinct package. Then, in theory, if you want to continue going down the frictionless route, the GUI could be `syncplay`, which would depend on e.g. `syncplay-common`/`-server`/`-cli`/whatever and the GUI dependencies. That would obviously be a quite heavy-handed approach, so I can understand if you're opposed.

PR #419, opened 2021-05-12
Allow playlists to be played from network

zeroeightysix:
`.m3u`, `.m3u8` (and probably txts as well) are frequently hosted on the web for TV stations, videos, and other media.

This PR checks if the "file path" starts with `http://` or `https://`, and, if so, does not try to load those 'files' from filesystem.

Previously syncplay would not be able to load videos from the internet through the `File > Open media stream URL` menu as it'd try to load them from the FS.

### DIFF
--- a/syncplay/client.py
+++ b/syncplay/client.py
@@ -569,7 +569,8 @@ class SyncplayClient(object):
         return False
 
     def openFile(self, filePath, resetPosition=False, fromUser=False):
-        if fromUser and filePath.endswith(".txt") or filePath.endswith(".m3u") or filePath.endswith(".m3u8"):
+        if not (filePath.startswith("http://") or filePath.startswith("https://"))\
+                and ((fromUser and filePath.endswith(".txt")) or filePath.endswith(".m3u") or filePath.endswith(".m3u8")):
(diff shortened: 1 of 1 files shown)


Et0h:
@zeroeightysix Would it be better to use the specific HTTP(S) check as proposed or the more generic utils.isURL protocol check which catches all *://* paths?

Et0h:
I'm a bit confused about the purpose of this PR - possibly because it doesn't give clear instructions on the functionality which it changes.

It is stated that previously Syncplay could not load videos through open media stream URL, but I tried loading a YouTube URL with the current version of Syncplay and the open media stream URL dialogue and it worked fine on mpv.

Et0h:
I tested the PR with mpv, mpv.net, VLC, MPC-HC (1.9.8) and it worked. As usual, streaming works best with Syncplay when using mpv/mpv.net due to how they handle seeks to areas which have yet to be buffered.

I used a Big Buck Bunny test stream:
`https://multiplatform-f.akamaihd.net/i/multi/will/bunny/big_buck_bunny_,640x360_400,640x360_700,640x360_1000,950x540_1500,.f4v.csmil/master.m3u8`

PR #434, opened 2021-06-26
Added room permanence option to server

Assistant:
Added the **--rooms-dir [directory]** option to the server.
This allows the specifying of a directory to be used to read and write room data in order for playlists to persist permanently, regardless of every watcher leaving or the server restarting.
The behavior won't take place if the directory does not exist, or if the **--isolate-rooms** option is enabled.

### DIFF
--- a/syncplay/server.py
+++ b/syncplay/server.py
@@ -421,11 +421,17 @@ class Room(object):
     def __str__(self, *args, **kwargs):
         return self.getName()
 
+    def roomsCanPersist(self):
+        return self._roomsDir is not None
+
+    def isPermanent(self):
+        return self.roomsCanPersist()
+
     def sanitizeFilename(self, filename, blacklist="<>:/\\|?*\"", placeholder="_"):
         return ''.join([c if c not in blacklist and ord(c) >= 32 else placeholder for c in filename])
 
     def writeToFile(self):
-        if self._roomsDir is None:
+        if not self.isPermanent():
             return
         if len(self._playlist) == 0:
             try:
@@ -487,7 +493,7 @@ class Room(object):
         return list(self._watchers.values())
 
     def addWatcher(self, watcher):
-        if self._watchers or self._roomsDir is not None:
+        if self._watchers or self.isPermanent():
             watcher.setPosition(self.getPosition())
         self._watchers[watcher.getName()] = watcher
         watcher.setRoom(self)
@@ -497,7 +503,7 @@ class Room(object):
             return
         del self._watchers[watcher.getName()]
         watcher.setRoom(None)
-        if not self._watchers and self._roomsDir is None:
+        if not self._watchers and not self.isPermanent():
             self._position = 0
         self.writeToFile()
 
--- a/syncplay/server.py
+++ b/syncplay/server.py
@@ -438,9 +438,10 @@ class Room(object):
     def writeToFile(self):
         if not self.isPermanent():
             return
+        filename = os.path.join(self._roomsDir, self.sanitizeFilename(self._name)+'.room')
         if len(self._playlist) == 0:
             try:
-                os.remove(os.path.join(self._roomsDir, self.sanitizeFilename(self._name)+'.room'))
+                os.remove(filename)
             except Exception:
                 pass
             return
@@ -450,7 +451,7 @@ class Room(object):
         data['playlistIndex'] = self._playlistIndex
         data['position'] = self._position
         data['lastSavedUpdate'] = self._lastSavedUpdate
-        with open(os.path.join(self._roomsDir, self.sanitizeFilename(self._name)+'.room'), "w") as outfile:
+        with open(filename, "w") as outfile:
             json.dump(data, outfile)
 
     def loadFromFile(self, filename):
--- a/syncplay/server.py
+++ b/syncplay/server.py
@@ -475,6 +475,7 @@ class Room(object):
         for watcher in self._watchers.values():
             watcher.setPosition(position)
             self._setBy = setBy
+        self.writeToFile()
 
     def isPlaying(self):
         return self._playState == self.STATE_PLAYING
@@ -486,7 +487,7 @@ class Room(object):
         return list(self._watchers.values())
 
     def addWatcher(self, watcher):
-        if self._watchers:
+        if self._watchers or self._roomsDir is not None:
             watcher.setPosition(self.getPosition())
         self._watchers[watcher.getName()] = watcher
         watcher.setRoom(self)
@@ -496,8 +497,9 @@ class Room(object):
             return
         del self._watchers[watcher.getName()]
         watcher.setRoom(None)
-        if not self._watchers:
+        if not self._watchers and self._roomsDir is None:
             self._position = 0
+        self.writeToFile()
 
     def isEmpty(self):
         return not bool(self._watchers)
--- a/syncplay/server.py
+++ b/syncplay/server.py
@@ -41,7 +41,7 @@ class SyncFactory(Factory):
             print(getMessage("no-salt-notification").format(salt))
         self._salt = salt
         self._motdFilePath = motdFilePath
-        self._roomsDirPath = roomsDirPath if os.path.isdir(roomsDirPath) else None
+        self._roomsDirPath = roomsDirPath if roomsDirPath is not None and os.path.isdir(roomsDirPath) else None
         self.disableReady = disableReady
         self.disableChat = disableChat
         self.maxChatMessageLength = maxChatMessageLength if maxChatMessageLength is not None else constants.MAX_CHAT_MESSAGE_LENGTH
--- a/syncplay/server.py
+++ b/syncplay/server.py
@@ -26,7 +26,7 @@ from syncplay.utils import RoomPasswordProvider, NotControlledRoom, RandomString
 
 
 class SyncFactory(Factory):
-    def __init__(self, port='', password='', motdFilePath=None, roomsDirPath=None, isolateRooms=False, salt=None,
+    def __init__(self, port='', password='', motdFilePath=None, roomsDirPath=None, roomsTimer=31558149, isolateRooms=False, salt=None,
                  disableReady=False, disableChat=False, maxChatMessageLength=constants.MAX_CHAT_MESSAGE_LENGTH,
                  maxUsernameLength=constants.MAX_USERNAME_LENGTH, statsDbFile=None, tlsCertPath=None):
         self.isolateRooms = isolateRooms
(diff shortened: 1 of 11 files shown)


syncplay/server.py
```diff
@@ -40,12 +41,13 @@ def __init__(self, port='', password='', motdFilePath=None, isolateRooms=False,
             print(getMessage("no-salt-notification").format(salt))
         self._salt = salt
         self._motdFilePath = motdFilePath
+        self._roomsDirPath = roomsDirPath if os.path.isdir(roomsDirPath) else None
```
Et0h, line 44: If --rooms-dir is not specified then this line results in the following error on my Windows 10 64-bit system:
```
File "\syncplay\syncplay\server.py", line 44, in __init__
    self._roomsDirPath = roomsDirPath if os.path.isdir(roomsDirPath) else None
TypeError: _isdir: path should be string, bytes or os.PathLike, not NoneType
```

syncplay/server.py
```diff
@@ -405,6 +421,27 @@ def __init__(self, name):
     def __str__(self, *args, **kwargs):
         return self.getName()
 
+    def writeToFile(self):
+        if self._roomsDir is None:
+            return
+        if len(self._playlist) == 0:
+            return
+        data = {}
+        data['name'] = self._name
```
Et0h, line 430: This input needs to be sanitized before it is used in a file path, or else someone could just set the room name to "..\foobar" to exit the folder constraints or use a `\n` character to make the server fail with a `OSError: [Errno 22] Invalid argument` error.

Et0h: Also, the use of the raw input means that rooms with a `:` in them don't get saved (at least on Windows 10), which is problematic as a colon is used in the names of managed rooms.

syncplay/server.py
```diff
@@ -405,6 +421,27 @@ def __init__(self, name):
...
+        with open(os.path.join(self._roomsDir, self._name+'.room'), "w") as outfile:
+            json.dump(data, outfile)
+
+    def loadFromFile(self, filename):
+        with open(filename) as json_file:
+            data = json.load(json_file)
+            self._name = truncateText(data['name'], constants.MAX_ROOM_NAME_LENGTH)
+            self._playlist = data['playlist']
+            self._playlistIndex = data['playlistIndex']
+            self._position = data['position']
```
Et0h, line 443: I've done a bit of testing and it seems like the position is not saved if people leave without pausing and when people join an empty room it doesn't correctly load the position either.

syncplay/server.py
```diff
@@ -405,6 +421,30 @@ def __init__(self, name):
...
         return self.getName()
 
+    def sanitizeFilename(self, filename, blacklist="<>:/\\|?*\"", placeholder="_"):
+        return ''.join([c if c not in blacklist and ord(c) >= 32 else placeholder for c in filename])
+
+    def writeToFile(self):
+        if self._roomsDir is None:
+            return
+        if len(self._playlist) == 0:
+            return
```
Et0h, line 431: If the playlist is set to blank then this should delete the playlist file rather than just return, or else it will never save it when people empty the playlist

Et0h:
I can see why it would be useful for those running private servers, but there might also be drawbacks which I have not thought of so I'll leave this open for comments to see what people think of this proposal before making any decisions as to whether to approve it in principle. It is possible that it might work best if this feature could be linked with a feature where a list of rooms which always show up in the room list can be specified (with the playlists being saved optionally being limited to this list).

Assistant:
I understand the concern about drawbacks—I sat on this code for about a year because originally it modified the behavior directly, and I didn't want to submit it until I refactored it to be optional, which I didn't have the motivation to do until now. However, besides that initial oopsie with the `roomsDirPath` argument validation, I believe the only difference between running this fork without the flag and the original should be a few functions running that immediately return.

I also can't overstate how useful it is for those of us running private servers that watch serial content with friends. Going from using regular syncplay to syncplay with persistent rooms was a QoL improvement on the same level as going from counting 3, 2, 1, go on irc to using syncplay. Being able to plan and setup viewing sessions ahead of time, not having to remember where you left off, not having to re-setup everything every time, being able to see what your group has watched together, and probably other minor improvements I'm not remembering.

Anyways, regardless of the outcome of this PR I'll continue working on this feature.

Assistant:
Also thanks a bunch for the feedback, it's greatly appreciated

daniel-123:
I feel like this would be pretty useful functionality for me personally and likely for quite a few other users.

From technical standpoint I'd prefer to see the persistent storage as SQLite database rather than a flat file with json in it. My reasoning is as follows:

- SQLite is a *very* robust library that allows for abstracting away most filesystem related problems - both about security and reliability.
- We already use SQLite in the server for basic statistics (Qt SQL from pySide), so there is no dependency growth.

From perspective of UX I think the default behavior should be that unused rooms either fade away (for example if nobody enters them for a month or a year) or/and get deleted if their playlist is empty. With those two it could even be possible to enable them along with room isolation (though I'm not convinced if that's a good combination).

I'm completely on the fence about whether there should be two types of persistence - "full" i.e. a list of rooms that never get deleted and "semi persistent" which would disappear under some specific conditions. This would be a bit complicated to convey properly in UI and easily perceived as inconsistent/buggy. Though if we assume that whoever controls the server also is an expert who can manage that as a niche feature it might make sense?

All that said - switching over to Qt SQL would require a lot of changes to the PR, so I'd understand if you weren't up for it. I also don't consider it a strict requirement before accepting the PR but rather an idea to think about.

Assistant:
I've made persistent rooms keep track of the time since last activity, and they get wiped if tried to interact with after a threshold (1 year by default, feel free to change this). It's configurable through `--rooms-timer`, and allows disabling with `0`.
Rooms without playlists are already deleted when the last user leaves, as of 5885a6ecf06095928a9b5b2402dba1576d28ee54.
I believe that should take care of the UX, although I'm not sure about isolated rooms, I don't use them so I don't understand them fully, so I didn't want to mess with them.

Regarding using a database, despite liking files from a sysadmin perspective as they're much easier to work with than a database, I do think a database would be best, however I'm not really up for it to be honest.

I think the idea in your edit is probably the best way to implement this feature, however I'm not confident enough to make a change like this to the default behavior of the program. If someone else wants to take a shot at it I think it'd be fantastic, though.

I also had the idea to make an option to disable creating new persistent rooms, only allowing rooms that already existed on boot to persist. What are your thoughts on that? Ideally I'd want to do something with the existing permissions system to allow operators to create new persistent rooms even in this limited state, but that's beyond my expertise and not something I'd use personally anyways.

tntmod54321:
This seems like it would be an excellent feature! +1 for adding it

Tremolo4:
I merged this feature into my private server and very much enjoy it, thanks @Assistant!

To add to the discussion: I think rooms should not be persistent by default and that there should be a way to explicitly set and unset them as persistent from the client, if the server allows it. That way one-time users don't accidentally leave stuff behind on the server.

Et0h:
Okay I've pushed an initial implementation of persistent/permanent rooms using sqlite to https://github.com/Syncplay/syncplay/tree/room_persistence

So far this only changes things server-side, and it works as follows:
* If `--rooms-db-file [path]` is specified then server-side persistent rooms will be enabled. This will create/use the specified sqlite database file to store the playlist and other room state information. Persistent rooms will be deleted if there are no users or playlist in them. Syncplay will list all persistent rooms in the list of rooms.
* If `--permanent-rooms-file [path]` is specified then it will load a textfile from that path with one permanent room per line. Permanent  rooms will always be listed and not delete even if it has no users of playlists in it.

The next step would be to provide a client-side way to toggle whether or not persistent rooms will be displayed (and to make them show up in the join comobox even when they are hidden).

Et0h:
I've been thinking about how to handle support for non-persistent (temporary) rooms for servers where persistent rooms are enabled. The approach I've just implemented is so that if you put -temp in a room name then it will make it a temporary room and not save the data and then use the MOTD system to inform users of how they need to use this feature if they want their room to be temporary.

This approach has the benefit of not requiring client UI changes which in turn makes it backwards-compatible. For those on the latest version of Syncplay the MOTD notice is added by the client which makes it use the client's language, but for older versions the server sends it in the server's language. Given that room names and MOTD messages are long-standing feature this means that it will work all the way back to the earliest versions of Syncplay.

As part of this process I noticed that the server didn't seem to use the system language for messages, so I've tried to fix this. @albertosottile can check to see if this broke anything in macOS as I had to tweak that code to get things to work on Windows.

...I'm hoping that this feature is now pretty much complete and ready for testing.

Et0h:
Okay, did a bit of testing the other day and it has a few bugs to fix before it is ready for testing. I'll post here when these are fixed.

Et0h:
Okay, it should be _actually_ ready to test now. Known issues:
- Currently no way for clients to filter out empty rooms.
- Things look fine in the GUI, but the userlist is a bit messy in CLI mode.

Et0h:
Updates:
* New GUI doesn't show blank users for empty rooms.
* New console UI don't see empty persistent rooms or get phantom userlist updates.
* The GUI window menu now has a 'Hide empty persistent rooms' option.

Are there any changes needed before this is considered ready for merging with the main branch?

Et0h:
Those running built versions of Syncplay can test the client/server at https://github.com/Syncplay/syncplay/releases/tag/v1.7.0-persistentrooms1

Tremolo4:
> Those running built versions of Syncplay can test the client/server at https://github.com/Syncplay/syncplay/releases/tag/v1.7.0-persistentrooms1

Is there a way to delete persistent rooms from the client? I can't seem to find one.

Et0h:
@Tremolo4 This is how it is supposed to work when persistent rooms are enabled:
* Temporary rooms exist so long as they have people in them. This is the same as how rooms operate without the persistent room feature enabled.
* Persistent rooms stick around so long as they have files in the playlist.
* Permanent rooms stick around even when they don't have files in the playlist.

If this is not how it is operating for you then please be more specific about what you are doing and what you mean by a room being 'saved'.

Tremolo4:
Thanks for the explanation!

I'm using 566f90b as the server, and [the release you posted](https://github.com/Syncplay/syncplay/releases/tag/v1.7.0-persistentrooms1) as the client.

Permanent rooms only exist if I supply a permanent-rooms-file to the server, right? I don't do that currently.

What I'm seeing right now is when I create any room at all, and then leave it again immediately, without adding anything to the playlist, it does not disappear from the room list. If I add ` -temp` to the name, it also does not disappear, until I restart the server.

The ` -temp` rooms also do not get saved to the sqlite database, unlike the other rooms. 👍

So it seems that it does not recognize the empty playlist for me? Maybe I'm doing something wrong still.

Et0h:
@Tremolo4 Okay it seems like something may have been broken as part of the merge process. I'll look into it.

Et0h:
@Tremolo4 I think I figured out what went wrong, which is that the room deletion code was erroneously condition on no db file being specified. Please let me know if 7083d8c fixes the bug.

Tremolo4:
That seems to have fixed it. Works flawlessly now for me so far.

Thanks a lot for your work on this feature!

PR #435, opened 2021-07-08
Console: Document setting offset in help

luke-jr:


Et0h:
I've not tested this yet, but this seems like a good improvement to me. Thanks for your contribution.

daniel-123:
Offset feature wasn't listed in there mostly because back in the day we thought it is confusing to use.

Thad said - I don't see a problem with adding it there, though in such case it should also be added to the man page in [docs/syncplay.1](https://github.com/Syncplay/syncplay/blob/master/docs/syncplay.1).

I can add it there myself after merging if you aren't up for it in this PR.

Et0h:
Do people still use this feature, and if so are they actually finding it worth the hassle? I find it not confusing and a lot more hassle than just ensuring that everyone has the same file, so I can appreciate what daniel-123 is saying about the reasons why it was not listed previously. However, there might still be people with slow Internet connections perhaps who find the feature useful.

Options include:
1. Keeping and documenting the feature, but adding some sort of warning in the documentation that its use is not recommended (i.e. that it is deprecated).
2. Removing the feature.
3. Keeping the feature, but making it a mostly undocumented to discourage use.

daniel-123:
Thinking about it a bit more:

- Offset is present in GUI anyway. I'd assume vast majority of users use GUI, so given it's visible there it makes sense for it to be documented as console option.
- Offset has gotten a bit more awkward with introduction of playlists. Right now it is set per instance of Syncplay running, but that doesn't translate well into watching multiple different files - not all of them might require offset so you need to manually keep track of it.

I don't see strong reasons to outright remove it (yet?), but marking it as depreciated along with documenting it sounds like best course of action for now.

@luke-jr given that you made the PR in first place, I assume you use the offset feature on at least somewhat regular basis. Could you share some feedback about how it works for you and how it's useful?

luke-jr:
Yes, my group uses it all the time. Often we have files that have variation in pre-content company names and such. Everyone getting the same file means waiting on slow downloads, and people who want super-high-quality (eg, 80 GB 4K) having to fall back to lower quality so those of us with pitiful bandwidth or smaller screens (scaling down doesn't work well) can actually watch.

We don't, however, use the playlist feature ever. It doesn't seem very useful - even though we watch multiple things in sequence sometimes, we typically don't plan what to watch next until the first one is over.

luke-jr:
(Please _don't_ deprecate the offset command! As mentioned, we rely on it)

Et0h:
> (Please _don't_ deprecate the offset command! As mentioned, we rely on it)

Your views have been noted.

Please be aware that within this context 'deprecate' does not mean remove, it means that we don't recommend using it and we don't promise to maintain support for it forever.

We wouldn't just remove support for it for no reason, but if we end up doing a refactor of the code to make way for a feature which cares about media position (such as if we ever allow for playback speed to be synced) we might decide to drop offset support to make it easier to implement that new feature.

PR #483, opened 2021-11-09
Windows dark mode

Et0h:


Et0h:
@sxiii @CalMaciborka @Zefferis@0lhi

Please test using dark mode support for Windows:
* Portable: https://github.com/Syncplay/syncplay/suites/4345878929/artifacts/114004663
* Installer: https://github.com/Syncplay/syncplay/suites/4345878929/artifacts/114004664

 Light/dark mode should follow your system settings for Windows 10 Anniversary users and above.

Et0h:
See #480 for a discussion on the title bar.

TeamDman:
Looks good to me!
Only comment is that the titlebar changes to white when unfocused, not sure if there's an easy solution to that which doesn't involve custom window controls like discord/spotify
image
image
image

Et0h:
@TeamDman This seems to relate to how Windows behaves rather than being a Syncplay-specific issue.

I just looked it up and following the advice at https://www.reddit.com/r/Windows10/comments/93e2yp/black_accent_color_in_windows_10/ worked for me. However, the advice is a bit old so I had to tweak what I did a bit: I had to set the 'Windows colours' under 'Colours' to a dark grey and also enable 'Title bars and window borders' under 'Show the accent colour on the following surfaces'. As usual, if you mess with your registry then you do so at your own risk.

PR #492, opened 2021-12-11
fix persistent rooms condition

odrling:
This commit fixes some conditions as seems to be intended after the changes in c0e8652c8c2a94f3a75d67bd0798292c55527e32.
Currently it crashes because it refers to attributes/methods that were removed in the mentioned commit.


Et0h:
Cheers for the fix.

Issue #355, opened 2020-10-02
PySide missing from AppImage of version 1.6.5


Sioma112233:
Downloading and running the AppImage for the latest version (1.6.5) on fedora 32 results in an error:

>> Could not import GUI libraries. If you do not have PySide installed then you will need to install it for the GUI to work.

This is unexpected as it's an AppImage distribution which I expected to have all dependencies included.
Installing the library (`PySide2`) to my system python does not help.

I also extracted the AppImage and ran a python shell using the "embedded python" in `./usr/bin/python` and could successfully `import PySide2`

Sioma112233:
I also found the module itself within the extract AppImage files under: `conda/lib/python3.7/site-packages/PySide2`

Sioma112233:
I've added a print of the exception just before [this line](https://github.com/Syncplay/syncplay/blob/16e33a4eef61c03e7b08b16dddbeefc4e7d3f1ec/syncplay/ui/ConfigurationGetter.py#L534)

it printed: 
>> cannot import name 'QtWidgets' from 'syncplay.vendor.Qt'

daniel-123:
I cannot replicate the issue (Debian 11) which is rather strange. I tried both official release of 1.6.5 and latest master for upcoming 1.6.6.

QtWidgets is part of PySide2, so I'm wondering what might be missing here. For the record I do have all the Qt libraries which Syncplay needs on my OS anyway as I normally use .deb installation.

That said - it *seems* like something that's specific either to Fedora 32 or even your system in particular. Otherwise I'd expect this to be reported way sooner. I'm not intimately familiar with our AppImage packaging process so I'm not quite certain how exactly it ensures all dependencies are packaged in. I know it uses requirements.txt and requirements_gui.txt, but those seem fine.

albertosottile:
I am also not able to reproduce on Ubuntu 18.04. Nevertheless, I think you are having an issue with [Qt.py](https://github.com/mottosso/Qt.py), the shim that we use to support both PySide and PySide2 at the same time. 

I would try to get some logs from this shim by using an environment variable designed for that. @Sioma112233, could you please open a terminal, go where you downloaded the AppImage file, and run
```bash
chmod +x Syncplay-1.6.5-x86_64.AppImage
export QT_VERBOSE=True
./Syncplay-1.6.5-x86_64.AppImage
```
and report here all the terminal output?

Thanks.

Sioma112233:
>> and report here all the terminal output?

[...]

PR #493, opened 2021-12-14
modernised python packaging

batmanfeynman:
I have made a pyproject.toml file as well as a setup.cfg file which is compliant with the latest python packaging recommendations.
These changes help me in creating an rpm for fedora as well. I will make an rpm package for v1.6.9 by including a patch file, but i think this modernized python packaging will help in possible future installation from pip.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,0 +1,61 @@
+[metadata]
+name = syncplay
+version = attr: syncplay.version
+author = Syncplay
+author_email = dev@syncplay.pl
+description = Client/server to synchronize media playback on mpv/VLC/MPC-HC/MPC-BE across many computers
+long_description = file: README.md
+long_description_content_type = text/markdown
+url = https://syncplay.pl/
+download_url = https://syncplay.pl/download/
+project_urls =
+    Source Code = https://github.com/Syncplay/syncplay
+    Bug Tracker = https://github.com/Syncplay/syncplay/issues
+license_file = LICENSE
+classifiers = 
+    Development Status :: 5 - Production/Stable
+    Environment :: MacOS X :: Cocoa
+    Environment :: Win32 (MS Windows)
+    Environment :: X11 Applications :: Qt
+    Framework :: Twisted
+    Intended Audience :: End Users/Desktop
+    License :: OSI Approved :: Apache Software License
+    Operating System :: MacOS :: MacOS X
+    Operating System :: Microsoft :: Windows
+    Operating System :: POSIX :: Linux
+    Natural Language :: English
+    Natural Language :: German
+    Natural Language :: Italian
+    Natural Language :: Russian
+    Natural Language :: Spanish
+    Programming Language :: Python :: 3
+    Programming Language :: Python :: 3.4
+    Programming Language :: Python :: 3.5
+    Programming Language :: Python :: 3.6
+    Programming Language :: Python :: 3.7
+    Programming Language :: Python :: 3.8
+    Programming Language :: Python :: 3.9
+    Programming Language :: Python :: 3.10
+    Programming Language :: Python :: 3 :: Only
+    Topic :: Internet
+    Topic :: Multimedia :: Video
+
+[options]
+include_package_data = True
+packages = find:
+install_requires =
+    twisted[tls]>=16.4.0
+    certifi>=2018.11.29
+    appnope>=0.1.0; sys_platform == 'darwin'
+    pypiwin32>=223; sys_platform == 'win32'
+    zope.interface>=4.4.0; sys_platform == 'win32'
+    pyside2>=5.11.0
+    requests>=2.20.0; sys_platform == 'darwin'
+python_requires = >=3.4
+
+[options.entry_points]
+console_scripts =
+    syncplay-server = syncplay.ep_server:main
+gui_scripts = 
+    syncplay = syncplay.ep_client:main
+
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,6 @@
+[build-system]
+requires = [
+    "setuptools>=51.0.0",
+    "wheel"
+]
+build-backend = "setuptools.build_meta"
--- a/setup.py
+++ b/setup.py
@@ -1,65 +1,6 @@
 #!/usr/bin/env python3
 
-import os
-import setuptools
+from setuptools import setup
+setup()
 
-from syncplay import projectURL, version as syncplay_version
 
-def read(fname):
-    with open(fname, 'r') as f:
-        return f.read()
-
-installRequirements = read('requirements.txt').splitlines() +\
-                        read('requirements_gui.txt').splitlines()
-
-setuptools.setup(
-    name="syncplay",
-    version=syncplay_version,
-    author="Syncplay",
-    author_email="dev@syncplay.pl",
-    description=' '.join([
-        'Client/server to synchronize media playback',
-        'on mpv/VLC/MPC-HC/MPC-BE on many computers'
-    ]),
-    long_description=read('README.md'),
-    long_description_content_type="text/markdown",
-    url=projectURL,
-    download_url=projectURL + 'download/',
-    packages=setuptools.find_packages(),
-    install_requires=installRequirements,
-    python_requires=">=3.4",
-    entry_points={
-        'console_scripts': [
-            'syncplay-server = syncplay.ep_server:main',
-        ],
-        'gui_scripts': [
-            'syncplay = syncplay.ep_client:main',
-        ]
-    },
-    include_package_data=True,
-    classifiers=[
-        "Development Status :: 5 - Production/Stable",
-        "Environment :: MacOS X :: Cocoa",
-        "Environment :: Win32 (MS Windows)",
-        "Environment :: X11 Applications :: Qt",
-        "Framework :: Twisted",
-        "Intended Audience :: End Users/Desktop",
-        "License :: OSI Approved :: Apache Software License",
-        "Operating System :: MacOS :: MacOS X",
-        "Operating System :: Microsoft :: Windows",
-        "Operating System :: POSIX :: Linux",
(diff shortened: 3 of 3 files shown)


daniel-123:
AppImage seems to work just fine on my Debian and this doesn't touch any of the deb packages, so from my side that seems to be covered.

One thing that isn't is that your new method assumes that end user will always want the GUI. Which is reasonably fair assumption on a client, but not so much in case of headless servers. Do you have some ideas how to have an optional set of requirements? Preferably, but not necessarily while keeping the GUI libraries as default.

batmanfeynman:
The changes introduced in #416 is what i came up with basically (those changes can easily be translated to the setup.cfg file). But that results in the following setup:
`pip install syncplay` would install everything without the gui requirements. 
`pip install syncplay[GUI]` would install the client gui as well.

i noticed that @FallenWarrior2k head searched for a way to make the GUI default and couldn't find one. I am inclined to believe them. But I will look around a bit more to see if it possible without any workarounds.

FallenWarrior2k:
I believe the extras-based approach is blocked on pypa/setuptools#1139.

batmanfeynman:
Also to add some information, I tried to see if i could use a venv with pip to run the syncplay code on both a fresh ubuntu and Fedora VM.
Ubuntu:
Installing requirements with pip went well and it fetched pyside2 5.15.2. All good. But syncplay client wouldn't run and complained about some missing library. I had to install some package through apt to get it working from within the venv

Fedora:
I made a venv. pip install for some reason only installed pyside2 5.13.1 . This was throwing some error on the python code of syncplay. Installing the python-packages from fedora repos did not get rid of this error! But now that all the dependencies were installed, if i exit the venv, syncplay would work.

The reason i am pointing out my experience is to show that getting gui working with pip based install method probably wont be straighforward.

batmanfeynman:
But I do agree with @FallenWarrior2k in that having at least the headless server package easily available on pypi would be awfully convenient.

Maybe we can make just the server package available through PyPI?

batmanfeynman:
@daniel-123 I also want to point out that with this commit, all i have done is translate the old setup.py file to the new format. The behaviour is the same, as in the old setup.py file also assumed that the end user will always want the GUI client.

What I want to point out is that making syncplay avaialable on PyPI is not the primary objective of this pull request, but instead to make it easier to follow packaging guidelines in fedora for projects developed in python. I will make a copr repo by tomorrow for 1.6.9 and link it here

albertosottile:
Before staring the review, just FYI, PySide2 wheels contain their own copy of Qt. So, in principle, there should be no need to resort to `apt` to run Syncplay after installing it from Python sources. However, there are some combinations of OS-PySide2 versions for which **other** libraries are missing when doing so. 

This is the case for Ubuntu 20.04.3 and PySide2 5.15.2, in which the library missing is `libxcb-xinerama.so.0` (see https://askubuntu.com/a/1069502). I just reproduced this on a fresh Focal Fossa VM. IMHO, this is a bug that should be reported upstream to PySide2. However, I can see that they cannot possibly support every target available... (though they should keep an eye on Ubuntu). 

In short, I would not consider impossible to install Syncplay with GUI from PyPi, provided that all the dependencies behave as they should. Nevertheless, I do not think that PyPI should be a supported distribution channel for this project, as we already have quite a few. We actually invest a lot of effort for providing binaries specifically prepared to avoid having to use the sources.

In addition, I do not see the benefits of PyPi over plain `git` for developers and enthusiasts that still want to run from sources, as Syncplay cannot be used as a dependency in other Python projects.

setup.cfg
```diff
@@ -0,0 +1,61 @@
...
+    Operating System :: Microsoft :: Windows
+    Operating System :: POSIX :: Linux
+    Natural Language :: English
+    Natural Language :: German
+    Natural Language :: Italian
+    Natural Language :: Russian
+    Natural Language :: Spanish
+    Programming Language :: Python :: 3
+    Programming Language :: Python :: 3.4
+    Programming Language :: Python :: 3.5
```
albertosottile, line 33: Same for the 3.4 and 3.5 classifiers, they should be deleted.

setup.cfg
```diff
@@ -0,0 +1,61 @@
...
+include_package_data = True
+packages = find:
+install_requires =
+    twisted[tls]>=16.4.0
+    certifi>=2018.11.29
+    appnope>=0.1.0; sys_platform == 'darwin'
+    pypiwin32>=223; sys_platform == 'win32'
+    zope.interface>=4.4.0; sys_platform == 'win32'
+    pyside2>=5.11.0
+    requests>=2.20.0; sys_platform == 'darwin'
```
albertosottile, line 53: Could you put these in alphabetical order?

setup.cfg
```diff
@@ -0,0 +1,61 @@
...
+    Programming Language :: Python :: 3.9
+    Programming Language :: Python :: 3.10
+    Programming Language :: Python :: 3 :: Only
+    Topic :: Internet
+    Topic :: Multimedia :: Video
+
+[options]
+include_package_data = True
+packages = find:
+install_requires =
```
albertosottile, line 46: I am not a fan of duplicated information, and switching to PEP 621 without touching the `requirements*.txt` files does exactly that. 

I would be in favor of deleting the requirement files, but currently the Windows build system relies on them. This could be circumvented by using `pip-compile` e.g. via
```
pip-compile --output-file requirements.txt setup.cfg
```
in the CI workflows before the file is actually needed.

On the other hand, we could retain the requirements files and start using them in the proper way: to pin dependencies for CI/release builds. We actually need to pin dependencies on macOS but, for the moment I did that in the workflow code (very bad, but I did not have a proper alternative up to now).

@Et0h @daniel-123 what is your opinion here?

albertosottile (changes requested):

setup.cfg
Et0h: > @Et0h @daniel-123 what is your opinion here?

As someone who is a Windows user I don't have any strong opinions on packaging beyond the general desire for things to work and be low-maintenance.

setup.cfg
```diff
@@ -0,0 +1,61 @@
...
+    twisted[tls]>=16.4.0
+    certifi>=2018.11.29
+    appnope>=0.1.0; sys_platform == 'darwin'
+    pypiwin32>=223; sys_platform == 'win32'
+    zope.interface>=4.4.0; sys_platform == 'win32'
+    pyside2>=5.11.0
+    requests>=2.20.0; sys_platform == 'darwin'
+python_requires = >=3.4
```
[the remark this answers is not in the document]
batmanfeynman: done

setup.cfg
batmanfeynman: done

setup.cfg
batmanfeynman: done

batmanfeynman:
Link to copr repo as promised:
https://copr.fedorainfracloud.org/coprs/batmanfeynman/syncplay/

I also want to suggest 3 modification to the syncplay.pl website:
1.  In [this page](https://syncplay.pl/guide/trouble/) under "Problems running Appimage of Syncplay on Fedora", the link to issue number #355 is not present. Please fix this.
2. Add the link to the copr page at the end of [this page](https://syncplay.pl/download/)
3. Maybe suggest using the rpm in my copr instead of the AppImage for now in the troubleshooting page

setup.cfg
daniel-123: Personally I'd prefer to avoid pinning dependencies to specific version unless strictly as rollback resulting from a bug in latest version of given dependency. At least as far as AppImage goes.

For OS packages - they just use what's on the OS, so pinning isn't really a useful concept in first place.

batmanfeynman: Hi @daniel-123 @albertosottile 
I just wanted to remind you of this!

soredake:
@batmanfeynman can you please add support for fedora 36 in your repo?

batmanfeynman:
@soredake done!

corbmr:
@batmanfeynman Would it possible to get fedora 37 added to your repo? Thank you for working on this!

batmanfeynman:
@corbmr  done.

setup.cfg
albertosottile: Sorry if I neglected this. The gist of this issue is that in both Windows and macOS workflows we currently have:
```
          pip3 install -r requirements.txt
          pip3 install -r requirements_gui.txt
```

and I do not see a clear way to reproduce the same behavior after this PR is merged UNLESS the information about dependencies is duplicated in `setup.cfg` and in the two requirements file. Am I missing something here? Is there a command to just install the dependencies of a package?

albertosottile:
General comment: the issue of having the GUI dependencies separate is not solved. I believe the vast majority of people that install from sources do that precisely because they do not want to have the GUI. Therefore, I do not think we can just merge this PR in a way that does not cover for this use case. Unless I am missing something, I think we need to keep the two separate requirements file, for the moment.

Et0h:
> Link to copr repo as promised: https://copr.fedorainfracloud.org/coprs/batmanfeynman/syncplay/
> 
> I also want to suggest 3 modification to the syncplay.pl website:
> 
> 1. In [this page](https://syncplay.pl/guide/trouble/) under "Problems running Appimage of Syncplay on Fedora", the link to issue number [PySide missing from AppImage of version 1.6.5 #355](https://github.com/Syncplay/syncplay/issues/355) is not present. Please fix this.

Fixed.

> 2. Add the link to the copr page at the end of [this page](https://syncplay.pl/download/)

I've added it to https://copr.fedorainfracloud.org/coprs/batmanfeynman/syncplay/ and https://github.com/Syncplay/syncplay/discussions/574

> 3. Maybe suggest using the rpm in my copr instead of the AppImage for now in the troubleshooting page

Done.

VarLad:
@Et0h @batmanfeynman Any idea if you could update the COPR Fedora 38 (and possibly build for Rawhide) and update the version to the latest release?

PR #496, opened 2022-01-10
Correct persistance argument type

modwizcode:
The current documentation specifies directories, likely as a result of the original design before use of a database, but the arguments expect files. This simply clarifies the documentation to match the code itself.


Et0h:
You're right, that was a legacy from an earlier iteration. Thanks for the fix!

PR #506, opened 2022-02-12
Fixed server argument max-username-length not working

powerjungle:


Et0h:
I can both confirm the bug and the fix. Well done for addressing this bug which I didn't even know existed.

PR #515, opened 2022-03-08
Fixes converting lines to array

danfai:
if condition was false convertMultilineStringToList returned "" instead
of an empty list.


Et0h (approved):

Et0h:
Nice find.

PR #516, opened 2022-03-08
Webhook API

danfai:
This PR adds a Webhook API to syncplay.
For example you can have a rocket-chat server, with channels that put every posted link in a channel in syncplay.
The API also works with persistent rooms and if somebody is in the channel it will post as someone randomly, since there is no user/watcher assigned to the API.
Note that currently this API does not allow reading the current playlist but only appending.

Example configuration for Outgoing WebHook integration in rocketchat:
URLs: http:// SYNCPLAY_SERVER : WEBHOOK_PORT
Script:
```
class Script {
  prepare_outgoing_request({ request }) {
    url = request.data.text;
    c = request.data.channel_name;
    request.data = {"text": text, "channel_name": c};
    return request;
    }
  }
​
  /**
   * @params {object} request, response
   */
  process_outgoing_response({ request, response }) {
    return;
  }
}
```

I am happy to hear your feedback and improve the PR if necessary.


Et0h:
Hi, this an interesting concept so thanks for sharing it.  It looks like it'd need some work to get it to be merge-ready and then there is the question of whether we want to add this level of additional complexity to the software.

As noted in the principles set out at https://github.com/Syncplay/syncplay/issues/315 we have to balance increasing functionality with keeping things as simple as possible to minimise bugs and make the software easier to maintain.

I'm not sure how many people would want every URL posted in their chat server to be added to a Syncplay playlist and webhooks isn't really an area I have that much familiarity with to be able to properly maintain and test the code. I'd be interested to hear from others what their thoughts are and whether or not it provides a feature which they would be likely to use.

dmgolembiowski:
> Hi, this an interesting concept so thanks for sharing it. It looks like it'd need some work to get it to be merge-ready and then there is the question of whether we want to add this level of additional complexity to the software.
> 
> As noted in the principles set out at #315 we have to balance increasing functionality with keeping things as simple as possible to minimise bugs and make the software easier to maintain.
> 
> I'm not sure how many people would want every URL posted in their chat server to be added to a Syncplay playlist and webhooks isn't really an area I have that much familiarity with to be able to properly maintain and test the code. I'd be interested to hear from others what their thoughts are and whether or not it provides a feature which they would be likely to use.

A webhook API seems to bring a lot of value here, but it might be better to move the webhook server out into its own package.

Et0h:
I'm closing this as we have no plans to support a Webhook API at present.

PR #518, opened 2022-03-22
Add /next command

rtix:
/next, /qn command to load next file in playlist. I find it useful for me, hence the proposal

### DIFF
--- a/syncplay/constants.py
+++ b/syncplay/constants.py
@@ -129,6 +129,7 @@ COMMANDS_QUEUEANDSELECT = ['queueandselect','qas']
 COMMANDS_PLAYLIST = ['playlist', 'ql', 'pl']
 COMMANDS_SELECT = ['select', 'qs']
 COMMANDS_DELETE = ['delete', 'd', 'qd']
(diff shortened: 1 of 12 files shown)


Et0h:
Nice work. I've tested it and it operates correctly. It's a simple change to do the code so shouldn't add any bugs, and I can see why it would be useful for those controlling Syncplay through mpv etc.

Unless anyone objects I'd say it's good to accept, with the remaining tasks being to:
* Add stubs for the "commandList-notification/next" message in other languages
* Add entry for the /next command to https://github.com/Syncplay/syncplay/blob/master/docs/syncplay.1
* Add entry to the manual on the website based on the explanations already provided.

rtix:
@Et0h I've added stubs, but I'm not sure how this command should look like in docs and how do I contribute to website?

Et0h:
> @Et0h I've added stubs, but I'm not sure how this command should look like in docs and how do I contribute to website?

Only I can update the website. I'll do the docs as well.

Et0h:
Just checked and at present manual is only for command line options, not /commands. As such I've just updated the website.

PR #521, opened 2022-04-04
Print timestamp when pausing.

Kansattica:
I would like to use Syncplay to watch a movie with a friend for a podcast where we watch together, pause to talk about things, and assemble it into a slide slow later. Syncplay printing when the media was paused in its log would help with this immensely.

### DIFF
--- a/syncplay/client.py
+++ b/syncplay/client.py
@@ -360,7 +360,7 @@ class SyncplayClient(object):
         self._player.setPaused(True)
         madeChangeOnPlayer = True
         if (self.lastLeftTime < time.time() - constants.OSD_DURATION) or hideFromOSD == True:
-            self.ui.showMessage(getMessage("pause-notification").format(setBy, self.getGlobalPosition()), hideFromOSD)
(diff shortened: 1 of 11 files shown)


Et0h:
This seems like a good idea to me - a small tweak that makes the notification more useful.

I'd be interested in people's thoughts on whether the proposed format of "X paused (00:05)" is better than going with "Foobar paused at 00:05". The latter is more in line with the format for "Foobar jumped from 14:04 to 12:56"  format used elsewhere and avoids conflict with "(MM:SS)" being the format used to display the duration when announcing "Foobar is playing 'Example.mkv' (23:40) in room: 'Videos'".

It could still use () for other languages for now as that doesn't require a translation, but we could add something to the stub that they should consider tweaking it to follow the English format.

Kansattica:
I originally wrote "Bob paused at 00:05" and changed it to the other phrasing because I didn't have the expertise to translate it into the other languages. If it's preferable to have a complete sentence for English and leave the others as stubs for now, I'm happy to make the change.

Et0h:
It's not a blocking issue so I'm happy to make the change myself now that I've confirmed you are happy with the wording.

Issue #553, opened 2022-10-08
SSL connection abruptly cut in certain countries


Someone7un7:
I can't connect to any of syncplay servers
What do I do?????

Et0h:
What operating system are you using? It is working fine for me using Windows 10, but https://github.com/Syncplay/syncplay/issues/554 reports someone having trouble connecting on Windows 11 due to a certificate error.

Et0h:
Please run Syncplay with the --debug command line switch and let me know of anything potentially relevant which gets put into the syncplay.log or .Syncplay.log file in %APPDATA%.

Et0h:
Also do you get any errors when you visit the https://syncplay.pl/ website in your browser?

Someone7un7:
no no I can simply visit syncplay.pl website without any error I mean I don't need to change my IP or something , and that --debug command which you said didn't do anything , by the way , I'm from Iran

Et0h:
This might be the same issue as https://github.com/Syncplay/syncplay/issues/554 and so might have the same solution.

Could you please try again with the build from https://github.com/Syncplay/syncplay/actions/runs/3235129755 and let me know if that fixes it.

Someone7un7:
> This might be the same issue as #554 and so might have the same solution.
> 
> Could you please try again with the build from https://github.com/Syncplay/syncplay/actions/runs/3235129755 and let me know if that fixes it.

I tried it and it didn't work , these are the notifications that are shown:

[17:30:02] Attempting to connect to syncplay.pl:21
[17:30:03] Attempting secure connection
[17:30:03] Successfully reached syncplay.pl (151.80.32.178)
[17:30:03] Connection with server lost, attempting to reconnect
[17:30:03] Attempting secure connection

I don't know why it doesn't work ☹
It was really a cool app

by the way, I have windows 10
tried everything buddy
I have no idea

maybe it is because of the internet problems that we have in Iran these days , but I don't think so because my friends can use syncplay , they can connect

Et0h:
@Someone7un7: With https://github.com/Syncplay/syncplay/pull/564 I've done some changes so that if an initial attempt fails for a TLS-related reason then it should now always try again using a plaintext connection but warn the user

Please try it out and let me know if you ca now use Syncplay: https://github.com/Syncplay/syncplay/actions/runs/3401531009 (scroll to the artifacts at the bottom and choose Syncplay-1.7.0-Setup.exe or Syncplay_1.7.0_Portable).

[...]

PR #564, opened 2022-11-05
Always try plaintext if TLS attempt fails

Et0h:


Et0h:
I'm hoping this addresses #553

Et0h:
I've been advised it would be better to track down the underlying cause than to change TLS behaviour.

PR #586, opened 2023-02-21
Log --debug as UTF-8 to fix charmap mapping

Et0h:
This should fix errors such as:
>  File "syncplay\players\mpv.pyc", line 743, in actuallySendLine
>     File "syncplay\client.pyc", line 1626, in showDebugMessage
> builtins.UnicodeEncodeError: 'charmap' codec can't encode character '\u3009' in position 684: character maps to <undefined>

When using Syncplay with mpv and--debug mode on Windows.

### DIFF
--- a/syncplay/utils.py
+++ b/syncplay/utils.py
@@ -195,13 +195,13 @@ def blackholeStdoutForFrozenWindow():
             _file = None
             _error = None
 
-            def write(self, text, fname='.syncplay.log', encoding='utf-8'):
+            def write(self, text, fname='.syncplay.log'):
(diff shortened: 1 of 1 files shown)


Et0h:
This seems to fix this issue which I experienced when running Syncplay with mpv when --debug was enabled. If nobody objects I'll merge it soon.

Context: The \u3009 symbol is a right-angle bracket, and it is the MPV_INPUT_PROMPT_START_CHARACTER specified in constants.py which is passed onto mpv to define the character which appears when you press enter to show you can write your message within mpv.

Issue #587, opened 2023-02-21
wildcard matching in trusted domains


name-snrl:
Wildcard matching doesn't work. I tried `trusteddomains = ['youtube.com', 'youtu.be', 'eu.ngrok.io']` and `trusteddomains = ['youtube.com', 'youtu.be', '.eu.ngrok.io']`.

Details here: _Originally posted by @name-snrl in https://github.com/Syncplay/syncplay/issues/437#issuecomment-1424710089_

Et0h:
@name-snrl &  @Tremolo4 : I've created a [pull request](https://github.com/Syncplay/syncplay/pull/588) which hopefully adds the feature requested here, but it uses regex so it's worth giving it lots of testing to make sure it doesn't break the trust logic in some weird way. Please give it a try via https://github.com/Syncplay/syncplay/actions/runs/4237707314 and let me know if it works via the pull request discussion.

Et0h:
The change has now been made. All that is left is to consider what, if any, changes want to be made to the client documentation at https://syncplay.pl/guide/client/

Et0h:
I've now updated the client guide, so this issue is being closed.

PR #588, opened 2023-02-21
Add wildcard matching of trusted subdomains (#587)

Et0h:
This change should mean that if someone specifies *.example.com as a trusted domain then abc.example.com will be recognised as a trusted domain. See #587 and #437.


name-snrl:
It works for me, thanks. As for safety, alas, I do not have sufficient competence in this matter

Et0h:
Thanks for testing it out. I wasn't expecting it to cause any problems, not least because it it only uses the regex if the wildcard is detected.

Et0h:
This is now included in Syncplay 1.7.0 RC 1: https://syncplay.pl/syncplay-1-7-0-release-candidate-1/

Issue #567, opened 2022-12-03
Shared playlist is empty for everyone after reconnection until something is added/removed in playlist


soredake:
**Describe the bug**
When you're alone in the room and you added something to shared playlist, after reconnecting to server shared playlist will be empty for everyone until you change shared playlist (add/remove something).

**To Reproduce**
1. Start syncplay, connect to room (you need to be alone in the room) and add something to playlist.
2. Disconnect from network.
3. After reconnection you will still see playlist items, but anyone who joins will see empty playlist until something is added/changed in playlist.

**Expected behavior**
Playlist should not be empty for everyone who will join up after the one who updated playlist reconnected to server.

**Screenshots**

**Version and platform:**
 - OS: Windows 11 Pro
 - Syncplay version and build type: Syncplay 1.7.0 beta 1
 - Media player and version:  latest mpv

**Additional context**
I've reproduced this in windows 11 <-> windows 11 with Syncplay 1.7.0 beta 1, with windows 11 <-> fedora kde 37 in vm with Syncplay 1.6.9 and with windows 11 <-> https://github.com/Reddnek/syncplay-android version 0.10.1

Et0h:
Thanks for the clarity of your bug report. It took me a bit of time to figure out a reasonable way to fix the issue by restoring the playlist when it is appropriate to do so (and not when it isn't), but I think I've got a working solution now.

@soredake Could you please test my proposed fix and and confirm that it resolves the issue: https://github.com/Syncplay/syncplay/actions/runs/4256136168

soredake:
@Et0h when i tested this build yesterday your fix is worked and playlist is restored after re-connection, but after retesting I've again see empty playlist in vm and syncplay-android, don't know why it's worked yesterday.

Et0h:
@soredake Thanks for the testing. Hopefully _this_ version should work: https://github.com/Syncplay/syncplay/actions/runs/4263298780

When testing, it is important to disconnect for at least 13 seconds to ensure that any 'ghost' version of the original connection have been timed out by the server.

soredake:
@Et0h disconnected for 5/20 seconds - can see playlist in vm/mobile app, thanks!

Et0h:
This tweak is now included in Syncplay 1.7.0 RC 1: https://syncplay.pl/syncplay-1-7-0-release-candidate-1/

soredake:
Thank you again for fixing this!

PR #590, opened 2023-02-23
Restore server-side playlist on reconnect (#567)

Et0h:
Address the issue raised in #567 where, if you were the only one in the room when you were disconnected, the shared playlist is empty for everyone else after reconnection until something is added/removed from the playlist.


Et0h:
This tweak is now included in Syncplay 1.7.0 RC 1: https://syncplay.pl/syncplay-1-7-0-release-candidate-1/

PR #591, opened 2023-03-04
Create messages_ko.py

MarongHappy:


Et0h:
Thanks for providing this translation. It would be helpful if you can confirm that you have tested these changes in both the installer and the client and that they work as expected. Build for this pull request can be found at: https://github.com/Syncplay/syncplay/actions/runs/4331596259

Et0h:
@MarongHappy Thanks for providing this translation. I've made a few tweaks to get it to work which are at https://github.com/Syncplay/syncplay/actions/runs/4407706296

It would be helpful if you can confirm that you have tested these changes in both the installer and the client and that they work as expected when using the Korean language mode. Can you please test: https://github.com/Syncplay/syncplay/actions/runs/4407706296

Et0h:
The translation is now included in Syncplay 1.7.0 RC 1: https://syncplay.pl/syncplay-1-7-0-release-candidate-1/

PR #592, opened 2023-03-13
Korean translation

Et0h:


Et0h:
The translation is now included in Syncplay 1.7.0 RC 1: https://syncplay.pl/syncplay-1-7-0-release-candidate-1/

PR #612, opened 2023-06-09
Added Embeded Chat Support For VLC Media Player

ixlammm:


Et0h:
Hi, this is a very interesting pull request and not something I was expecting. It is really quite ingenious and quite a feat.

Part of me wants to ask you to finish it off by:
* Adding support for other operating systems
* Making it build the via the build system
* Making it something which people can toggle in options via the Chat tab
* Seeing if you can add/confirm support for non-standard key characters to better support non-English languages

Another part of me thinks that maybe this is something that is not right for Syncplay because:
* I don't fully understand this code and don't know C programming, so would not be able to maintain it or fix bugs.
* We want to encourage people to move over to mpv rather than give them more reasons to stick with VLC as mpv is generally more reliable for Syncplay.
* When you first launch VLC using this it appeared to give me a weird command line prompt privilege message. This might be a red flag for some users, and might make Syncplay more likely to be picked up by virus scanners (a problem we are already facing).

...So for now I'll reserve my position and invite further comments while I consider things.

ixlammm:
Hey, I originally wrote the code on Ubuntu but then I moved to windows and the code didn't change that much so getting in to run on other operating systems wouldn't be that hard. I think adding support for non-ASCII characters would be possible since it's using osd to display the messages.
For someone who's been using vlc for 10+ years, mpv isn't as good as vlc and I'd rather stick with it.
The reason you got the prompt privilege message is because I has to copy the DLL file somehow and I chose to do it in python because I didn't know how to do it in the installation.
After all, this is not a project that I wanted to spend a lot of time in, It was just for fun.
Sorry for the unprofessional response.

daniel-123:
Hi @ixlammm 

Besides long term support which you already discussed, I'll also make a side comment for future reference - including a binary dll file in the repository is not good in terms of security. See recent [XZ vulnerability](https://en.wikipedia.org/wiki/XZ_Utils_backdoor), which was introduced by a malicious actor in similar way.

Another aspect is that the code you committed doesn't actually include information how to compile it. Syncplay uses continuous integration, so the expected thing would be to include entire process of building this dll file straight into it - so that its latest version can be automatically included in the generated installer packages etc.

PR #638, opened 2023-10-14
Use ansi escape and readline so output doesnt mix with input in cli

gospodin55:
It works on linux, tell me if I should handle Windows console differently

### DIFF
--- a/syncplay/ui/consoleUI.py
+++ b/syncplay/ui/consoleUI.py
@@ -4,6 +4,7 @@ import sys
 import threading
 import time
 import os
+import readline
 
 import syncplay
 from syncplay import constants
@@ -105,10 +106,15 @@ class ConsoleUI(threading.Thread):
             message = message.decode('utf-8')
         except UnicodeEncodeError:
             pass
+        sys.stdout.write('\33[2K\r')
         if noTimestamp:
             print(message)
         else:
             print(time.strftime(constants.UI_TIME_FORMAT, time.localtime()) + message)
+        line = readline.get_line_buffer()
+        if line != '':
+            print(line, end='')
+            sys.stdout.flush()
 
     def showDebugMessage(self, message):
         print(message)


Et0h:
In Windows I get the following error:
`Traceback (most recent call last):
  File "syncplayClient.py", line 14, in <module>
  File "<frozen zipimport>", line 259, in load_module
  File "syncplay\ep_client.pyc", line 3, in <module>
  File "<frozen zipimport>", line 259, in load_module
  File "syncplay\clientManager.pyc", line 2, in <module>
  File "<frozen zipimport>", line 259, in load_module
  File "syncplay\ui\__init__.pyc", line 14, in <module>
  File "<frozen zipimport>", line 259, in load_module
  File "syncplay\ui\consoleUI.pyc", line 7, in <module>
ModuleNotFoundError: No module named 'readline'`

I assume that this is because readline has not been added to the buildPy2exe include list, i.e.:
https://github.com/Syncplay/syncplay/blob/20846ca78ccdeb07dc7529445fde2868a3ce6c37/buildPy2exe.py#L682

Et0h:
I think the reason it errors on Windows is because readline needs to be added to requirements.txt, but I could be wrong.

gospodin55:
It seems there's no native readline implementation for windows, I think it's too much to handle, I will just check for the current system and disable this feature if readline is unavailable

Et0h:
Ah, okay.

Et0h:
Just checked, and it looks like if there was demand for that functionality in the future it might be possible to do so with pyreadline3, but nobody has raised any concerns about the current behaviour of SynplayConsole.exe so for now it doesn't seem worth the complexity.

daniel-123:
I've tried to test this on Debian 13 - with this change closing the application (either by ctrl+c or by closing the media player) results in basically non-functional prompt. As in typing commands further doesn't cause them to show up for example, ctrl+c adds two new lines etc.

Since I'm not quite sure what exactly in detail is wrong with current behavior of Syncplay and what's the expected result of this pull request would actually do - only feedback I can currently provide is that this doesn't seem good. Could you explain what the "output doesnt mix with input in cli" means in an example?

gospodin55:
image
If a user receives a message while they're typing, it appears on the same line as their input and interferes with it.
I don't have the same issue as you. Maybe that depends on the shell/terminal, I should've tested it more

daniel-123:
I'm using largely standard Debian, so this is just bash.

What distro/terminal did you use?

PR #645, opened 2023-11-09
Use PySide6 for non-windows + python >= 3.12

luk1337:
PySide2 is dead and won't support Python 3.12 or newer.

### DIFF
--- a/requirements_gui.txt
+++ b/requirements_gui.txt
@@ -1,3 +1,3 @@
-pyside2>=5.11.0; sys_platform != 'darwin'
-pyside6; sys_platform == 'darwin'
+pyside2>=5.11.0; sys_platform != 'darwin' and (sys_platform == 'win32' or python_version < '3.12')
+pyside6; sys_platform == 'darwin' or (sys_platform != 'win32' and python_version >= '3.12')
 requests>=2.20.0; sys_platform == 'darwin'


luk1337:
Looks like Windows build fails, should I do:
```diff
diff --git a/requirements_gui.txt b/requirements_gui.txt
index ba12db7..208cb3b 100644
--- a/requirements_gui.txt
+++ b/requirements_gui.txt
@@ -1,3 +1,3 @@
-pyside2>=5.11.0; sys_platform != 'darwin'
-pyside6; sys_platform == 'darwin'
+pyside2>=5.11.0; sys_platform == 'win32'
+pyside6; sys_platform != 'win32'
 requests>=2.20.0; sys_platform == 'darwin'
```
instead, or attempt to migrate win32 to PySide6?

luk1337:
> In terms of which PySide to use, the constraints are as follows:
> 
> * PySide6 doesn't work on Windows build of Syncplay (as we use the x86 version of Python 3.8 rather than the x64 version)
> * We want to continue to use PySide6 for macOS
> * Packaging of PySide6 on several different Linux distros is still not done
> 
> Taking this into account, for maximum compatibility the GUI requirements should be PySide6 for darwin (macOS) and for non-Darwin:
> 
> * PySide2 for Python < 3.12
> * PySide6 for Python >= 3.12
> 
> We might move Windows to PySide6 and 64-bit in the future, but probably best not to make that switch without extensive testing.

Does this look fine?

````
pyside2>=5.11.0; sys_platform != 'darwin' and (sys_platform == 'win32' or python_version < '3.12')
pyside6; sys_platform == 'darwin' or (sys_platform != 'win32' and python_version >= '3.12')
requests>=2.20.0; sys_platform == 'darwin'
```

Et0h:
> Does this look fine?

From a visual inspection it looks good to me, although I am no requirements.txt expert. I guess the next thing to do is to make the commit on your branch for the build to update and then we test if it works in practice.

luk1337:
> > Does this look fine?
> 
> From a visual inspection it looks good to me, although I am no requirements.txt expert. I guess the next thing to do is to make the commit on your branch for the build to update and then we test if it works in practice.

Ok, I force pushed my branch. Also I tried `pip install . --force` on Linux and it seemed fine.

Et0h:
Runs fine for me on Windows 10.

PR #646, opened 2023-11-10
Remove broken setAlignment() calls

luk1337:
Qt.AlignTop & Qt.AlignLeft is 0, so we might as well not call setAlignment() at all.


Et0h (approved):
I've tested this change and the changes don't seem to break anything on Windows 10.

daniel-123:
I've tested the build on Debian Linux and it doesn't seem to break anything.

PR #647, opened 2023-11-10
Don't pass empty model to QCompleter

luk1337:
It's empty by default.


daniel-123:
Tested on Debian Linux with no issues observed.

PR #649, opened 2023-11-26
Fixed German Translation and added more Translated Strings

maxwai:
Fixed the message that is shown if the client doesn't support TLS (it was saying the server doesn't support TLS).

Also added not yet done translation.


Et0h:
Cheers, thanks for this work.

PR #662, opened 2024-01-05
Add support for mpv installer from microsoft store

soredake:
Tested in windows 11 vm


Et0h:
I tested and it works fine on Windows 10. The command primpt issue you highlight might be a Windows 11 only issue.

Does setting Syncplay to send mpv the `--no-terminal` player argument resolve the issue for you on Windows 11 VM? If so, it might want to add that as a workaround for that specific use case.

Issue #665, opened 2024-01-22
Cannot use spaces or quotes in player arguments for vlc


zuluwi:
When I enter the following code to specify the snapshot folder in the player arguments, it gives an error due to spaces even though the folder path is in quotation marks

```
--language en --extraintf=luaintf:http --http-port=12342 --http-password=vlcpassword --snapshot-path "D:\stok\test folder\snapshots"
```
![resim](https://github.com/Syncplay/syncplay/assets/111116092/f21cf031-80fe-42af-a583-4883f1154580)

also the arguments where I use quotation marks without spaces don't work. In cmd I can run the following code properly
```
"C:\Program Files\VideoLAN\VLC\vlc.exe" --extraintf=luaintf:http --http-port=12342 --http-password=vlcpassword --snapshot-path "D:\stok\test folder\snapshots" --snapshot-format="jpg" --snapshot-prefix="image-" 
```

Is this a bug or am I not entering these characters properly?

**Version and platform:**
 - OS: Windows 10 Pro 22H2 64-bit
 - Syncplay version and build type: Syncplay 1.6.9 portable
 - Media player and version: VLC 3.0.20 64-bit

Et0h:
I've had a look and I think it's a bug. It looks like if you want a space in a snapshot path that for now you will need to either manually set it in VLC preferences or use mpv/mpv.net.

zuluwi:
Thanks for your reply! I specifically wanted to be able to specify the snapshot location by entering arguments. By the way, when I try with mpv, I get the same result, in the `syncplay.ini` file the folder path seems to be separated even though there are quotation marks

```ini
perplayerarguments = {'c:\\program files\\videolan\\vlc\\vlc.exe': ['--extraintf=luaintf:http', '--http-port=12342', '--http-password=vlcpassword', '--snapshot-path="D:\\stok\\test', 'folder\\snapshots"', '--snapshot-format="jpg"', '--snapshot-prefix="image-"', '--drawable-hwnd=1'], 'd:\\stok\\test folder\\snapshots\\mpv.exe': ['--screenshot-dir="D:\\stok\\test', 'folder\\snapshots"']}
```

Et0h:
Just checked and you're right, it is incorrectly splitting stuff in syncplay.ini.

Et0h:
I've tried to fix the issue in PR #668

[...]

PR #668, opened 2024-01-26
Allow player arguments with spaces/quotes (#665)

Et0h:
Hopefully this will fix the VLC issue #665 raised by @zuluwi and also work for other players.


Et0h:
Successfully tested on VLC, mpv and mpv.net.

Issue #529, opened 2022-05-13
[mpv] Overrides `input-ipc-server` set in mpv.conf


ahmubashshir:
**Describe the bug**
Syncplay overrides `input-ipc-server` set in `mpv.conf` thus breaking tools that depends on known `input-ipc-server`

**To Reproduce**
1. Set `input-ipc-server=/run/user/1000/mpv` in mpv.conf
2. Run mpv with syncplay

**Expected behavior**
Syncplay should check mpv.conf and use ipc path from mpv.conf if it's set.

**Version and platform:**
 - OS: Arch Linux
 - Syncplay version and build type: 1.6.9-5 [community/syncplay]
 - Media player and version:  mpv 1:0.34.1-4 [community/mpv]

edit: replace outdated option name.

Et0h:
I hope you can find a solution to this problem, but unfortunately it is not something I can be of much assistance with as it falls outside my knowledge and use case.

Syncplay uses [JSON IPC](https://github.com/iwalton3/python-mpv-jsonipc) to communicate with [mpv](https://github.com/mpv-player/mpv/), and in my understanding JSON IPC sets `input-ipc-server` in mpv to allow for this communication. As such, it is possible that the problem / solution lies 'upstream' rather than with Syncplay.

If you can figure out a way for JSON IPC to add itself to input-ipc-server without overriding input-ipc-path then feel free to make a pull request to the appropriate repositories. It's possible that liaising with JSON IPC / mpv devs would be of assistance.

ahmubashshir:
This patch defines a new function that returns either the ipc path
set in mpv.conf or the default(temporary) ipc path.

Closes #529

Signed-off-by: Mubashshir ***@***.***>
---
 requirements_gui.txt    |  1 +
 syncplay/players/mpv.py | 26 +++++++++++++++++++++++++-
 2 files changed, 26 insertions(+), 1 deletion(-)

diff --git a/requirements_gui.txt b/requirements_gui.txt
index a51f665..738e3ba 100644
--- a/requirements_gui.txt
+++ b/requirements_gui.txt
@@ -1,2 +1,3 @@
 pyside2>=5.11.0
 requests>=2.20.0; sys_platform == 'darwin'
+appdirs>=1.4.4; sys_platform != 'darwin'
diff --git a/syncplay/players/mpv.py b/syncplay/players/mpv.py
index a5cd06d..7968aec 100755
--- a/syncplay/players/mpv.py
[...]

PR #675, opened 2024-04-07
Fix MPV socket getting created in CWD

### LINKED ISSUE #674: mpvSyncplaySocket created in $HOME
**Describe the bug**
Since 459c9e91985796210fbc36968f8a6e572141f24f, a file named `mpvSyncplaySocket` is created in my home folder.

**Expected behavior**
On Linux, these kind of files would normally be created under `$XDG_RUNTIME_DIR`.
This env var might be undefined on non-systemd setups, in which case it should probably default to `$TMPDIR` (and fall back to `/tmp` if undefined).

Or as the [XDG Base Directory spec](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) puts it:
> `$XDG_RUNTIME_DIR` defines the base directory relative to which user-specific non-essential runtime files and other file objects (such as sockets, named pipes, ...) should be stored.

You can find more information about the standard file hierarchy (as defined by systemd) here: https://man7.org/linux/man-pages/man7/file-hierarchy.7.html

**Version and platform:**
 - OS: Arch Linux
 - Syncplay version: [Syncplay 1.7.2](https://archlinux.org/packages/extra/any/syncplay/)
 - Media player and version:  [MPV 0.37.0](https://archlinux.org/packages/extra/x86_64/mpv/)

notpeelz:
Fixes #674

I haven't tested on Windows or MacOS.

### DIFF
--- a/syncplay/players/mpv.py
+++ b/syncplay/players/mpv.py
@@ -621,8 +621,7 @@ class MpvPlayer(BasePlayer):
                     env['PATH'] = python_executable + ':' + env['PATH']
                     env['PYTHONPATH'] = pythonPath
             try:
-                socket = self.mpv_arguments.get('input-ipc-server')
-                self.mpvpipe = self.playerIPCHandler(mpv_location=self.playerPath, ipc_socket=socket, loglevel="info", log_handler=self.__playerController.mpv_log_handler, quit_callback=self.stop_client, env=env, **self.mpv_arguments)
+                self.mpvpipe = self.playerIPCHandler(mpv_location=self.playerPath, loglevel="info", log_handler=self.__playerController.mpv_log_handler, quit_callback=self.stop_client, env=env, **self.mpv_arguments)
             except Exception as e:
                 self.quitReason = getMessage("media-player-error").format(str(e)) + " " + getMessage("mpv-failed-advice")
                 self.__playerController.reactor.callFromThread(self.__playerController._client.ui.showErrorMessage, self.quitReason, True)
--- a/syncplay/players/mpv.py
+++ b/syncplay/players/mpv.py
@@ -375,7 +375,7 @@ class MpvPlayer(BasePlayer):
         self._listener.sendLine(['loadfile', filePath], notReadyAfterThis=True)
 
     def setFeatures(self, featureList):
-        self.sendMpvOptions()
+        self._sendMpvOptions()
 
     def setPosition(self, value):
         if value < constants.DO_NOT_RESET_POSITION_THRESHOLD and self._recentlyReset():
@@ -408,7 +408,7 @@ class MpvPlayer(BasePlayer):
             self._storePosition(0)
         # TO TRY: self._listener.setReadyToSend(False)
 
-    def sendMpvOptions(self):
+    def _sendMpvOptions(self):
         options = []
         for option in constants.MPV_SYNCPLAYINTF_OPTIONS_TO_SEND:
             options.append("{}={}".format(option, self._client._config[option]))
@@ -420,8 +420,9 @@ class MpvPlayer(BasePlayer):
         options_string = ", ".join(options)
         self._listener.sendLine(["script-message-to", "syncplayintf", "set_syncplayintf_options",  options_string])
         self._setOSDPosition()
-        publicIPCSocket = self._listener.mpv_arguments.get("input-ipc-server") if self._listener.mpv_arguments.get("input-ipc-server") else "mpvSyncplaySocket"
-        self._setProperty("input-ipc-server", publicIPCSocket)
+        socketPath = self._listener.mpv_arguments.get("input-ipc-server")
+        if socketPath is not None:
+            self._setProperty("input-ipc-server", socketPath)
 
     def _handleUnknownLine(self, line):
         self.mpvErrorCheck(line)
@@ -449,7 +450,7 @@ class MpvPlayer(BasePlayer):
             #self._client.ui.showDebugMessage("{} = {} / {}".format(update_string, paused_update, position_update))
 
         if "<get_syncplayintf_options>" in line:
-            self.sendMpvOptions()
+            self._sendMpvOptions()
 
         if line == "<SyncplayUpdateFile>" or "Playing:" in line:
             self._client.ui.showDebugMessage("Not ready to send due to <SyncplayUpdateFile>")
--- a/syncplay/players/mpv.py
+++ b/syncplay/players/mpv.py
@@ -1,5 +1,6 @@
 # coding:utf8
 import os
+import random
 import re
 import sys
 import time
@@ -10,7 +11,7 @@ import ast
 from syncplay import constants
 from syncplay.messages import getMessage
 from syncplay.players.basePlayer import BasePlayer
-from syncplay.utils import isURL, findResourcePath
+from syncplay.utils import getRuntimeDir, isURL, findResourcePath
 from syncplay.utils import isMacOS, isWindows, isASCII
 from syncplay.vendor.python_mpv_jsonipc.python_mpv_jsonipc import MPV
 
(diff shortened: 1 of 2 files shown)


syncplay/utils.py
```diff
@@ -37,9 +39,27 @@ def isMacOS():
 def isBSD():
     return constants.OS_BSD in sys.platform or sys.platform.startswith(constants.OS_DRAGONFLY)
 
+
 def isWindowsConsole():
     return os.path.basename(sys.executable) == "SyncplayConsole.exe"
 
+
+def getRuntimeDir():
+    if cachedPath := getattr(getRuntimeDir, "cachedPath", None):
```
notpeelz, line 48: I just realized the walrus operator is a Python 3.8+ thing. I'll change this to something that works on older versions of Python.

Et0h: That would be good because Syncplay is currently set up to support Python 3.6 and above.

notpeelz: Oh? It says Python >=3.4 in `setup.py` and `syncplayClient.py`

syncplay/players/mpv.py
```diff
@@ -408,7 +408,13 @@ def openFile(self, filePath, resetPosition=False):
             self._storePosition(0)
         # TO TRY: self._listener.setReadyToSend(False)
 
-    def sendMpvOptions(self):
+    def _getMpvSocketPath(self):
+        val = self._listener.mpv_arguments.get("input-ipc-server")
+        if val is not None:
+            return val
+        return getRuntimeDir().joinpath("mpv-socket").as_posix()
```
notpeelz, line 415: Windows expects a named pipe identifier, not a a path, e.g. `\\.\pipe\mpvsocket`

Reference: https://github.com/mpv-player/mpv/blob/2d4fae4f706c5c400acbc8a27d7819b4d539a315/DOCS/man/ipc.rst#command-prompt-example

@Et0h should I use a fully-qualified named pipe here?

notpeelz: It should probably also be unique, in case the user runs multiple instances of syncplay simultaneously.

notpeelz: I updated the patch. Let me know if it works for you (I don't have a Windows setup to test with atm).

syncplay/utils.py
Et0h: > Oh? It says Python >=3.4 in `setup.py` and `syncplayClient.py`

Oops, it looks like those files are out of date in that regard (along with `syncplayServer.py`).

https://syncplay.pl/guide/install/ states that on Linux / *BSD we expect people to use python >= 3.6 and that the last version which supported Python 3.4 is Syncplay 1.6.5 (which was released back in June 2020).

I can't remember what broke Python 3.4 support, but our compatibility is generally tied to PySide compatibility.

The Windows .exe is built using Python 3.8, the macOS dmg uses Python 3.7.4, and the deb package code says it relies on python >= 3.6.

syncplay/players/mpv.py
Et0h: > It should probably also use a unique random identifier, in case the user runs multiple instances of syncplay simultaneously.

The whole point of the default name was that it is something predictable, not random as per https://github.com/Syncplay/syncplay/issues/529#issuecomment-1923722267 

If people have multiple versions of Syncplay we already have a way to allow people to specify the socket name for instance which is to pass --input-ipc-server=socketname as a player argument within Syncplay.

Et0h: > On Windows, MPV expects a named pipe identifier (not a a path), e.g. `\\.\pipe\mpvsocket`
> 
> Reference: https://github.com/mpv-player/mpv/blob/2d4fae4f706c5c400acbc8a27d7819b4d539a315/DOCS/man/ipc.rst#command-prompt-example
> 
> @Et0h should I use a fully-qualified named pipe here?

I could be wrong, but I thought this was done by the JSON IPC Bridge: https://github.com/Syncplay/syncplay/blob/0a9700058e4b65676fb53567dee4ff781eddf955/syncplay/vendor/python_mpv_jsonipc/python_mpv_jsonipc.py#L52

notpeelz: > The whole point of the default name was that it is something predictable, not random as per https://github.com/Syncplay/syncplay/issues/529#issuecomment-1923722267

If the user wants a predictable name, they should specify it explicitly. A randomized name will prevent weird crashes from happening when running multiple instances.

> If people have multiple versions of Syncplay we already have a way to allow people to specify the socket name for instance which is to pass --input-ipc-server=socketname as a player argument within Syncplay.

My changes don't undo that. That should still work the same.

notpeelz: > I could be wrong, but I thought this was done by the JSON IPC Bridge:
> 
> https://github.com/Syncplay/syncplay/blob/0a9700058e4b65676fb53567dee4ff781eddf955/syncplay/vendor/python_mpv_jsonipc/python_mpv_jsonipc.py#L52

Good catch, I'll fix it.

Et0h: > If the user wants a predictable name, they should specify it explicitly. A randomized name will prevent weird crashes from happening when running multiple instances.

A randomized name is no use as there is no way for third party applications to know what that name is other than making an educated guess.

If you are aware of evidence of weird crashes being caused by mpv trying to use an IPC socket that already exists then that could be a reason to just not set up a secondary IPC port unless someone explicitly sets it via --input-ipc-server=socketname.

I don't see anything in the mpv docs warning against using an `input-ipc-server `name that might already be in use. https://github.com/mpv-player/mpv/blob/2d4fae4f706c5c400acbc8a27d7819b4d539a315/input/ipc-unix.c#L334 seems to indicate that if it can't bind the socket it just logs an error and aborts the socket creation process, but I'm not that familiar with the mpv code. The equivalent Windows code might be https://github.com/mpv-player/mpv/blob/2d4fae4f706c5c400acbc8a27d7819b4d539a315/input/ipc-win.c#L422 but it's less clear.

Et0h: > Edit: upon further inspection, it seems like the `\\.\pipe\xxx` prefix is only prepended when syncplay connects to the named pipe. MPV receives `--input-ipc-server=xxx` (no `\\.\pipe\` prefix).

I think mpv has its own code for that which is https://github.com/mpv-player/mpv/blob/2d4fae4f706c5c400acbc8a27d7819b4d539a315/input/ipc-win.c#L470

notpeelz: > A randomized second socket name is no use as there is no way for third party applications to know what that name is other than making an educated guess.

There's a problem with using a hardcoded socket path:

```
$ mpv --input-ipc-server=~/socket video.mp4 &
# pause the player 
$ echo 'cycle pause' | socat - ~/socket

# whoops, we end up clobbering the socket file...
$ mpv --input-ipc-server=~/socket video.mp4 &
# ...now this controls the second player and the first socket is in limbo
$ echo 'cycle pause' | socat - ~/socket
```

Besides, if the user wants to do IPC with MPV, they should explicitly pass the `--input-ipc-server` option, otherwise the socket name/path is just an implementation detail between Syncplay and MPV.

notpeelz: This seems like a bug:

With an explicit `--input-ipc-server=blah` option, you get duplicated args (i.e. `--input-ipc-server=blah --input-ipc-server=\\.\pipe\blah`):
image

With the default, you get a single arg (i.e. `\\.\pipe\mpvXXXXXXXXX`):
image

notpeelz: I've updated the PR.

With the default (no extra player args):
```
/usr/bin/mpv ... --input-ipc-server=/run/user/1000/syncplay-0vx3hrff/mpv-socket
```

With a user-defined `--input-ipc-server=/tmp/mpvtest` option:

```
/usr/bin/mpv ... --input-ipc-server=/tmp/mpvtest --input-ipc-server=/run/user/1000/syncplay-5vk5d3kg/mpv-socket
```

daniel-123:
Overall I think the PR looks quite reasonable with one caveat: the python_mpv_jsonipc.py is technically vendored from https://github.com/iwalton3/python-mpv-jsonipc 

It would be best if Syncplay didn't add too much of its own things to jsonipc, but on the other hand I don't think anybody else uses that library to begin with and we already made some changes to it. So it's not like this is a blocker of any kind, just something to be aware of and documented in this PR.

notpeelz:
I could pass the socket path via the `ipc_socket` constructor parameter instead and leave the vendored library unmodified.

daniel-123 (approved):

Et0h (approved):
I have tested in Windows by confirming that mpv works and by using using Power Shell and `[System.IO.Directory]::GetFiles("\\.\\pipe\\")` to confirm that by default it will create a pipe along the lines of `\\.\\pipe\\syncplay-mpv-94244417144407` but that if you add a player argument such as `input-ipc-server=\\.\pipe\foobar` it will also create a pipe of the specified name.

Et0h:
Thanks @notpeelz and @daniel-123 for your work on this. Time to accept those changes! :-)

PR #685, opened 2024-06-24
Polish translation

GitesHubisz:
Hello. I have managed to translate syncplay to polish. It was tiring, but I think it will do something. Pls proof-check is everything fine. Thanks in advance.

### DIFF
--- a/docs/syncplay.1
+++ b/docs/syncplay.1
@@ -97,7 +97,7 @@ Path to the player binary.
 
 .TP
 .B \-\-language language
-Language of Syncplay messages. Valid values are de/en/es/eo/fi/fr/it/pt_PT/pt_BR/tr/ru/zh_CN/ko.
+Language of Syncplay messages. Valid values are de/en/es/eo/fi/fr/it/pl_PL/pt_PT/pt_BR/tr/ru/zh_CN/ko.
 
 .TP
 .B \-\-clear\-gui\-data
--- a/syncplay/messages.py
+++ b/syncplay/messages.py
@@ -8,6 +8,7 @@ from . import messages_eo
 from . import messages_fi
 from . import messages_fr
 from . import messages_it
+from . import messages_pl_PL
 from . import messages_pt_PT
 from . import messages_pt_BR
 from . import messages_tr
@@ -25,6 +26,7 @@ messages = {
     "fi": messages_fi.fi,
     "fr": messages_fr.fr,
     "it": messages_it.it,
+    "pl_PL": messages_pl_PL.pl_PL,
     "pt_PT": messages_pt_PT.pt_PT,
     "pt_BR": messages_pt_BR.pt_BR,
     "tr": messages_tr.tr,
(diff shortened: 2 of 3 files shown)


Et0h:
Hi, thanks for all the work you put into this. There really are quite a lot of messages to translate, and some of the messages are quite technical.

I'll let those who speak Polish review the quality of the translation. However, I can give two pieces of feedback on the translation from a technical perspective.

Firstly the value of Language in messages_pl_PL.py is "Polish" which is the English name for the language, but as per https://github.com/Syncplay/syncplay/wiki/Guide-to-translating-Syncplay-into-another-language the style is for the language to be written in the local name for the language. This would presumably be Polszczyzna or polski. Is there a reason why you called it Polish, or was this just a case where you thought we wanted the English name for the language?

Secondly, you translated `chatdirectinput-label` text of `Allow instant chat input (bypass having to press enter key to chat)` to `Zezwalaj na natychmiastowe wprowadzanie wiadomości do czatu (omiń konieczność naciskania klawisza Enter, aby czatować)`

This is a very lengthy translation, which significantly increases the width of the configuration window.

Is there a shorter way to say this that would allow for the window to be narrower? For example, something like e.g. `Natychmiastowe wprowadzanie wiadomości do czatu (omiń naciskania klawisza Enter)` or `Natychmiastowe wprowadzanie wiadomości (bez konieczności naciskania klawisza Enter)` would be a reasonable length.

GitesHubisz:
Thanks for the answer. I say up front that it's been a while since I've done this, so some things I'll have to remind myself of, but I'm gonna try to answer:

> Is there a reason why you called it Polish, or was this just a case where you thought we wanted the English name for the language?

And the reason is that I am used to keeping English translation of that word and I just forgot to check in the wiki/other translations.

> Zezwalaj na natychmiastowe wprowadzanie wiadomości do czatu (omiń konieczność naciskania klawisza Enter, aby czatować)
> This is a very lengthy translation, which significantly increases the width of the configuration window.

and

Is there a shorter way to say this that would allow for the window to be narrower? For example, something like e.g. Natychmiastowe wprowadzanie wiadomości do czatu (omiń naciskania klawisza Enter) or Natychmiastowe wprowadzanie wiadomości (bez konieczności naciskania klawisza Enter) would be a reasonable length.

I tried to keep the same scheme to all message, because I think changing during translation would leave a "bad taste". Maybe changing the whole scheme would be a option, but I guess it depends now on Daniel. 

But as for your proposals, both have missing "allow" in it, so "zezwalaj" should stay. Maybe 
`Zezwalaj na natychmiastowe wprowadź wiadomość do czatu (omiń wciskanie klawisza Enter)` would fit.

I jeszcze Daniel; mam nadzieje, że nie wyrwiesz włosów podczas sprawdzania. :P

przestudent:
looking forward to the polish translation. Willing to help in any way if you'd allow

GitesHubisz:
I have nothing against it

PR #687, opened 2024-07-04
Update Turkish translation

muaz742:
Each #translate tagged line has been translated.


Et0h:
Thank you for this these translations.

PR #689, opened 2024-07-15
Adjust QSplitter style

luk1337:
This looks much better when using PySide6 6.7.2.

### DIFF
--- a/syncplay/ui/gui.py
+++ b/syncplay/ui/gui.py
@@ -1510,6 +1510,8 @@ class MainWindow(QtWidgets.QMainWindow):
         if isMacOS(): window.userlistLayout.setContentsMargins(3, 0, 3, 0)
 
         window.listSplit = QtWidgets.QSplitter(Qt.Vertical, self)
+        window.listSplit.setHandleWidth(6)
+        window.listSplit.setStyle(QtWidgets.QStyleFactory.create("fusion"))
         window.listSplit.addWidget(window.userlistFrame)
         window.listLayout.addWidget(window.listSplit)
         window.roomsCombobox = QtWidgets.QComboBox(self)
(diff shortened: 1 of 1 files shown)


luk1337:
Before:
Screenshot_20240715_180447

After:
Screenshot_20240715_180439

luk1337:
I guess it'd be great if someone could test pyside2 and other platforms.

Et0h:
Tried on Windows 10 dark mode, and the difference is that the old version had a clear grab indicator for the horizontal splitter but the new version does not (both have a grab indicator for the vertical handle). I didn't see any difference for the light mode (where there are no vertical/horizontal grab bars in any case).
Old and new grab bar Syncplay

I wouldn't say having horizontal but not vertical grab bars on Windows 10 dark mode is an improvement, as it means the way it displays it is inconsistent between horizontal and vertical. I might not even notice if I wasn't paying attention, but making an explicit change which is a downgrade on Windows doesn't seem like an improvement. I don't know what OS and theme you are using where it works better for you, but it might also be the case that it looks worse in other circumstances as well.

luk1337:
> Tried on Windows 10 dark mode, and the difference is that the old version had a clear grab indicator for the horizontal splitter but the new version does not (both have a grab indicator for the vertical handle). I didn't see any difference for the light mode (where there are no vertical/horizontal grab bars in any case). Old and new grab bar Syncplay
> 
> I wouldn't say having horizontal but not vertical grab bars on Windows 10 dark mode is an improvement, as it means the way it displays it is inconsistent between horizontal and vertical. I might not even notice if I wasn't paying attention, but making an explicit change which is a downgrade on Windows doesn't seem like an improvement. I don't know what OS and theme you are using where it works better for you, but it might also be the case that it looks worse in other circumstances as well.

hmm, I tried Window 10 in light mode and there was no indicator, so I just killed it. But if dark mode somehow has it, I guess I'll maybe try to look into where it comes from.

luk1337:
What about current diff? It looks like this on my system:

image

luk1337:
And looks like this without custom handle width:
image

Et0h:
The new version looks fine to me on Windows 10 Dark Mode now.

daniel-123:
Over here with pySide2 it seems to look exactly the same as originally.

PR #690, opened 2024-07-15
Fix broken macOS check

luk1337:


luk1337:
Before:
Screenshot_20240716_005940

After:
Screenshot_20240716_005931

selurvedu:
@luk1337 you posted the same screenshot twice.

luk1337:
> @luk1337 you posted the same screenshot twice.

no, look at the highlighted part:
image

selurvedu:
Oooh, riiight. That's so hard to notice if you don't know where to look. Sorry for the noise. :sweat_smile:

luk1337:
> Oooh, riiight. That's so hard to notice if you don't know where to look. Sorry for the noise. 😅

np, you can also notice that SSL icon has different style ^^

Et0h:
The fix looks good on Windows, but we have not tested it on macOS and the first change seems to be targeted at that OS. Have you tested your first change on macOS to confirm it fixes an issue rather than introducing one?

luk1337:
> The fix looks good on Windows and Linux, but we have not tested it on macOS and the first change seems to be targeted at that OS. Have you tested your first change on macOS to confirm it fixes an issue rather than introducing one?

ok, I tested it on macOS and this was breaking it, so I just restored it to how it was before.

```
if isMacOS():
    window.outputlabel.setMinimumHeight(21)
else:
    window.outputlabel.setMinimumHeight(27)
```

And with just fixing broken macOS check, this change now only affects non-macOS.

luk1337:
> > The fix looks good on Windows and Linux, but we have not tested it on macOS and the first change seems to be targeted at that OS. Have you tested your first change on macOS to confirm it fixes an issue rather than introducing one?
> 
> ok, I tested it on macOS and this was breaking it, so I just restored it to how it was before.
> 
> ```
> if isMacOS():
>     window.outputlabel.setMinimumHeight(21)
> else:
>     window.outputlabel.setMinimumHeight(27)
> ```
> 
> And with just fixing broken macOS check, this change now only affects non-macOS.

image

Et0h:
Thanks for the additional testing and fixes. The updated version seems good to go! 👍

Issue #408, opened 2021-04-08
m3u playlist parsing


GreatGrogrodor:
I've noticed when i create an m3u8 playlist, Syncplay doesn't parse metadata in the list. Not a huge deal, but it can make things more readable when your username and passwords that are in those links are 128 character randomly generated strings lol

Example.m3u8:
```m3u
#EXTM3U
#EXTINF:123, Movie1
https://username:password@domain.com/Movies/some_movie.mp4
#EXTINF:123, Movie2
https://username:password@domain.com/Movies/some_other_movie.mp4
```

VLC media player will display the names of the movies as "Movie1" and "Movie2" that are 2 minutes and 3 seconds long (123 seconds) rather than the URL, if i open it directly. However syncplay does this:
![image](https://user-images.githubusercontent.com/17716439/114063471-8a6c5480-9890-11eb-9bf4-be92aee21257.png)

It's treating the comments/metadata in the file as though it's a movie to play, rather than the title info & other metadata.

So, i have some suggestions. Either:
A: Make the playlist a "DataGridView Control" or at least that's what it's called in winforms, though i see you're using qt, so whatever the equivalent is in there. Add columns for Artist, Album, Title, Track number, URL, Duration, etc. Have it populate that list from the metadata/URLs.

B: Make it ignore empty lines and metadata. It'll still pull the links through though and it'll work. It just won't be purdy.

I'm willing to put in a herculian effort to add this myself someday, but i have a lot on my plate right now, and probably will for a while. Any questions on implementation, I'm happy to answer though.

There's some documentation here on the subject:
https://docs.fileformat.com/audio/m3u/

Et0h:
I am not in favour of making the playlist more complex as per the principles set out at https://github.com/Syncplay/syncplay/issues/315

However, I think it would be an improvement to strip comments etc from m3u files would be happy to see a PR for this. I think that to properly parse arbitrary m3u files Syncplay might also need to strip out local full and relative paths (if it doesn't already do this). At present Syncplay is only really designed to parse the m3u8 files created through the Save Playlist option (so only a subset of what the format allows).

Et0h:
Okay, the more I look into the details of the .m3u format the more I want Syncplay to simply support txt files with one filename/character per line. Issue include handling filenames starting with a # and using special characters. I'm now in favour of just changing the default file format .txt and making it clear in our documentation that Syncplay does not provide full m3u support.

Et0h:
[...]

PR #695, opened 2024-08-31
Update gui.py

0fbcb238c0:
Update browserfilter to allow saving playlists as m3u8 by default.

### DIFF
--- a/syncplay/ui/gui.py
+++ b/syncplay/ui/gui.py
@@ -1137,7 +1137,7 @@ class MainWindow(QtWidgets.QMainWindow):
             defaultdirectory = currentdirectory
         else:
             defaultdirectory = self.getInitialMediaDirectory()
-        browserfilter = "Playlist (*.txt)"
+        browserfilter = "Playlist (*.txt *.m3u8)"
         filepath, filtr = QtWidgets.QFileDialog.getSaveFileName(
             self, "Save playlist to file", defaultdirectory,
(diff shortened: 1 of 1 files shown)


Et0h:
Syncplay does not actually support .m3u8 because we do not support the  metadata this format can contain. The only reason we still had loading for it is because we used to allow people to load/save in the .m3u8 name but without it supporting the standard and as part of #408 back in 2021 we removed the saving but kept the loading in case someone had previously saved the playlist. We entitled this as "Begin move from m3u/m3u8 to txt for playlist" as it was only intended to be a temporary measure.

Given that we have no plans to support loading true m3u8 files and it seems to be causing confusion it might be best to remove the load m3u8 option.

0fbcb238c0:
> Syncplay does not actually support .m3u8 because we do not support the metadata this format can contain. The only reason we still had loading for it is because we used to allow people to load/save in the .m3u8 name but without it supporting the standard and as part of #408 back in 2021 we removed the saving but kept the loading in case someone had previously saved the playlist. We entitled this as "Begin move from m3u/m3u8 to txt for playlist" as it was only intended to be a temporary measure.
> 
> Given that we have no plans to support loading true m3u8 files and it seems to be causing confusion it might be best to remove the load m3u8 option.

Alright, good to know. I will close this issue and change the extension of my playlist to txt.
Thank you

PR #697, opened 2024-09-28
Migrate to actions/upload-artifact@v4

luk1337:
https://github.blog/changelog/2024-02-13-deprecation-notice-v1-and-v2-of-the-artifact-actions/


Et0h:
Good move - seems like an updated was needed 👍

PR #705, opened 2024-12-10
Allow `@` keyind to be used (mpv Next Chapter)

Assistant:
This allows the default keybind for Next Chapter in mpv to be used when the option to automatically chat is enabled.

### DIFF
--- a/syncplay/resources/syncplayintf.lua
+++ b/syncplay/resources/syncplayintf.lua
@@ -965,7 +965,6 @@ add_alpharowbinding('[',']')
 add_alpharowbinding('#','#')
 add_alpharowbinding('~','~')
 add_alpharowbinding('\'','\'')
-add_alpharowbinding('@','@')
 
 add_specialalphabindings(non_us_chars)
 add_repl_bindings(bindings)


Et0h:
Thanks for this contribution. I use mpv, but I don't use the instant chat feature. It wold be nice to hear from at least one other person who would be affected by this change about what they think the expected/ideal behaviour is.

Assistant:
Of course.
For context the current behavior is if you press `@` instead of skipping to the next chapter it would open the chat and type "@" to it.
With this change pressing `@` will skip to the next chapter, while allowing "@" to be typed if the chat is opened first.
I've been using this patch for a long time so I don't remember if it's possible to use the keybind bypassing the instant chat, but if it is I believe it wasn't very obvious.

Et0h:
> I've been using this patch for a long time so I don't remember if it's possible to use the keybind bypassing the instant chat, but if it is I believe it wasn't very obvious.

According to https://syncplay.pl/guide/client/ "In some cases when in the ‘instant chat’ mode you might actually want to trigger one of the keys from the alphabet rows, such as Control+T to set set 'stay on top' to Yesand Control+Shift+T to set it to No. To do this you first have to press ‘Tab’ to enter into the ‘Alphabet rows temporarily behave as normal’ mode. You press tab again to exit that mode."

Pressing tab is also mentioned in the tooltip for the option within the configuration and within mpv input screen when the feature is enabled. It also tells you within mpv to press tab to let you know you have temporarily disabled it and to press tab again to return to instant chat mode.

Assistant:
Reading might not be my strong suit.
Regardless, personally the case where I want to go to the next chapter is a lot more frequent than the case where I want to start a chat message with "@", so I think that should be the easier option.

Et0h:
Before coming to a decision I'll ask some people listed in the Syncplay change logs as contributing to our mpv support in the past two years to give them an opportunity to share their perspective if they wish to do so: @mbalandis @notpeelz @soredake @ahmubashshir @ducreyna @hunbernd @Abu-AM

mbalandis:
I personally don't use neither @ to skip chapters nor the chat feature so it wouldn't affect me (probably) however I can't speak for others... and there is a small risk imho. I think the best practice is to add an option in settings (use @ for mpv to skip chapters). Should be relatively simple instead of just hardcoding it like in this PR. Again... might be fine as it is but this is how as a dev would do it.

Et0h:
> I personally don't use neither '@' to skip chapters nor the chat feature (very rarely use it) so it wouldn't affect me (probably) however I can't speak for others... and there is a small risk imho. I think the best practice is to add an option in settings (use '@' for mpv to skip chapters). Should be relatively simple instead of just hardcoding it like in this PR. Again... might be fine as it is but this is how as a dev would do it.

Best practice from one angle, but doing that would arguably increase the complexity of the settings dialogue for something which won't matter to most people. One way to handle it is remove the @ keybinding but also note in the changelog that people should post on this issue if they want us to make it optional.

gavtroy:
I'd start a line with `@name` to direct it at someone, or even drop `@name` on its own to jokingly tag them in a character on screen. Meanwhile a lot of users won't know `@` is a shortcut in the first place and if it accidentally skips forward for no apparent reason that's a lot more disruptive than if it accidentally opens chat.

I'm also used to `@` for skipping chapter, but when it hasn't even been listed in the manual for over 10 years (https://github.com/mpv-player/mpv/commit/c4feba847c4cdabf036b4f2a8e55fa3b04c71412) I don't think it deserves an exception.

PR #706, opened 2024-12-21
Lossless image compression

Awqre:
Lossless optimization of PDF, ICNS and ICO files and all PNG files for file size reduction

Combination of different optimizations (ECT, PNGOut, DeflOpt...)


Et0h:
This is an interesting effort that seems to have the most impact on the deb files:
* Reduces file size by around 670 KB, mostly from icons.icns (453 KB) and to a lesser extent icon.ico (77 KB) and to an even lesser extent the larger syncplay.png files.
* Does not alter size of Syncplay-1.7.4-Setup.exe, which remains 19.5 MB
* Shrinks Syncplay_1.7.4_Portable from 27 MB to 26.9 MB
* Shrinks both syncplay-server.deb and syncplay.deb from 1.5 MB to 1.08 MB

I have two questions:
* Other than file size, is the re-encoding expected to have any positive and/or negative impact on performance?
* What platforms (and Python version if relevant) have you tested to ensure that these changes don't cause any compatibility issues?

Awqre:
The optimizations shouldn't have any negative performance impacts. And, due to lower file size it should load faster on any system but may not be noticeable, it will almost certainly not affect memory usage either way but if it does it would be in a positive way.

The PNG optimizations are lossless and within the format specifications so they should be compatible with any standard usage on any OS. Only incompatibilities I expect would be if for some reason there's a specific requirement for an image's palette order or bit depth to remain unchanged.

I've tested on both MacOS and Windows 11, both seem to work fine but I wasn't extensive. Windows XP and older will not work with PNG ICO files. If you find any issues on Linux systems I would be interested in testing different methods for compatibility.

Et0h:
Thanks for the additional information. Syncplay does not officially support operating systems that have reached End of Life status from their developers (for a variety of practical reasons) and so a lack of Windows XP support is not an issue.

Awqre:
> 3. What commands or procedure would I need to follow if I wished to reproduce this work for new graphics? For example, which command line arguments did you use and did you use a tool to automate the process?

I use a collection of programs to get better compression, I'll try to list them all out later but you can get good results using a program called FileOptimizer or running ZopfliPNG and/or OxiPNG. (I'm on the train so I'll edit with links soon)

Awqre:
> > 3. What commands or procedure would I need to follow if I wished to reproduce this work for new graphics? For example, which command line arguments did you use and did you use a tool to automate the process?
> 
> I use a collection of programs to get better compression

Many of these programs are windows only, closed source, and for some, the creators have disappeared or passed away.

My set:
[PNGOut](https://advsys.net/ken/utils.htm)
[OxiPNG](https://github.com/shssoichiro/oxipng)
[ZopfliPNG](https://github.com/google/zopfli)
[ECT](https://github.com/fhanau/Efficient-Compression-Tool)
[deft4j](https://github.com/NeRdTheNed/deft4j)
[DeflOpt](https://encode.su/threads/3257-Deflopt)

[FileOptimizer](https://sourceforge.net/projects/nikkhokkho/files/FileOptimizer/) has most of the ones I use plus more, I usually run it before my own:
[PngOptimizer](https://psydk.org/pngoptimizer)
[TruePNG](http://x128.ho.ua/pngutils.html)
[Leanify](https://github.com/JayXon/Leanify)
[pngwolf](https://github.com/hoehrmann/pngwolf)
[pngrewrite](https://entropymine.com/jason/pngrewrite/)
[advpng](https://github.com/amadvance/advancecomp)
[pingo](https://css-ig.net/pingo)
[defluff](https://encode.su/threads/1214-defluff-a-deflate-huffman-optimizer)

Et0h:
Thanks for that additional clarification. Merging now to save some space!

PR #707, opened 2024-12-23
Fix invocation of get font dialog

ritiek:
Fixes the following traceback when opening the get font dialog through Chat -> Set Font:
```python
  File "/home/username/syncplay/syncplay/ui/GuiConfiguration.py", line 1030, in <lambda>
    self.chatInputFontButton.released.connect(lambda: self.fontDialog("chatInput"))
                                                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/username/syncplay/syncplay/ui/GuiConfiguration.py", line 1131, in fontDialog
    font.setWeight(self.config[configName + "FontWeight"])
TypeError: 'PySide6.QtGui.QFont.setWeight' called with wrong argument types:
  PySide6.QtGui.QFont.setWeight(float)
Supported signatures:
  PySide6.QtGui.QFont.setWeight(PySide6.QtGui.QFont.Weight)
```

### DIFF
--- a/syncplay/ui/GuiConfiguration.py
+++ b/syncplay/ui/GuiConfiguration.py
@@ -1128,7 +1128,7 @@ class ConfigDialog(QtWidgets.QDialog):
             font = QtGui.QFont()
             font.setFamily(self.config[configName + "FontFamily"])
             font.setPointSize(self.config[configName + "RelativeFontSize"])
-            font.setWeight(self.config[configName + "FontWeight"])
+            font.setLegacyWeight(self.config[configName + "FontWeight"])
             font.setUnderline(self.config[configName + "FontUnderline"])
             ok, value = QtWidgets.QFontDialog.getFont(font)
             if ok:


Et0h:
Thanks for trying to fix this issue in GuiConfiguration.py. Before it is accepted it needs to be tested to ensure it works across the variety of set-ups supported by Syncplay

What platform and combinations of Python / PySide have you tested this on?

Also, do the following lines from gui.py also need updating?
https://github.com/Syncplay/syncplay/blob/d6de53e11eadb1da0f76e6d21873ecaac8e93882/syncplay/ui/gui.py#L626
https://github.com/Syncplay/syncplay/blob/d6de53e11eadb1da0f76e6d21873ecaac8e93882/syncplay/ui/gui.py#L696

ritiek:
> What platform and combinations of Python / PySide have you tested this on?

I was able to reproduce this issue on my EndeavourOS Linux with Python 3.12.3 and also on MacOS 14.2 Sonoma with Python 3.13.0, both of which seem to use PySide6 by default when syncplay is installed from source (d6de53e11eadb1da0f76e6d21873ecaac8e93882).

> Also, do the following lines from gui.py also need changing to use setLegacyWeight or are they fine as they are?
> 
> https://github.com/Syncplay/syncplay/blob/d6de53e11eadb1da0f76e6d21873ecaac8e93882/syncplay/ui/gui.py#L626
> 
> 
> https://github.com/Syncplay/syncplay/blob/d6de53e11eadb1da0f76e6d21873ecaac8e93882/syncplay/ui/gui.py#L696

These seem to be fine as they are.

Issue seems to be that `setWeight` in PySide6 no longer accepts int/float anymore (which is what we've been passing through `self.config[configName + "FontWeight"]`), and only accepts a `Weight` object now. Whereas the `setLegacyWeight` in PySide6 still accepts int/float values.
https://doc.qt.io/qtforpython-6/PySide6/QtGui/QFont.html#PySide6.QtGui.QFont.Weight
https://doc.qt.io/qtforpython-6/PySide6/QtGui/QFont.html#PySide6.QtGui.QFont.setLegacyWeight

I think the change so far will break syncplay environments that depends on PySide2 as PySide2 doesn't have the `setLegacyWeight` method. PySide2's `setWeight` accepts int/float, so we can maybe add a check to call the appropriate method depending on what PySide version is being used.
https://doc.qt.io/qtforpython-5/PySide2/QtGui/QFont.html#PySide2.QtGui.PySide2.QtGui.QFont.setWeight

Et0h:
@ritiek This longstanding issue should hopefully be fixed by https://github.com/Syncplay/syncplay/pull/795

PR #709, opened 2024-12-31
Special-case MOTD in showMessage procedures.

addap:
Hello, for a MOTD it is natural to use ASCII art but since messages in the chat box are added using QT's HTML-like text fragments, multiple spaces are merged together and the default variable-width font destroys alignment.

<table>
<tr>
<td>default look:</td>
<td>desired look:</td>
</tr><tr><td>

broken

</td><td>

correct

</td></tr></table>

Therefore, I propose that when showing a MOTD we escape spaces with `&nbsp;` and wrap everything in `<code>` tags to select a monospace font.

One problem is that MOTD is also used for showing warnings to the user, such as the "new-syncplay-available-motd-message" message. These will also be changed accordingly. If undesired, these warning should be sent in a separate field.

I have only tested this on openSUSE with python 3.11.11 and PySide6 6.8.0.2 so far. I assume that `<code>` tags work in every python QT library since they are supported by the rich text engine (https://doc.qt.io/qtforpython-5/overviews/richtext-html-subset.html#supported-html-subset). Is there a standard way to test for the different QT wrapper libraries that should be supported?

### DIFF
--- a/syncplay/client.py
+++ b/syncplay/client.py
@@ -1687,10 +1687,10 @@ class UiManager(object):
     def setSSLMode(self, sslMode, sslInformation=""):
         self.__ui.setSSLMode(sslMode, sslInformation)
 
-    def showMessage(self, message, noPlayer=False, noTimestamp=False, OSDType=constants.OSD_NOTIFICATION, mood=constants.MESSAGE_NEUTRAL):
+    def showMessage(self, message, noPlayer=False, noTimestamp=False, OSDType=constants.OSD_NOTIFICATION, mood=constants.MESSAGE_NEUTRAL, isMotd=False):
         if not noPlayer:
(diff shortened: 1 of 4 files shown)


Et0h:
Hi, sorry for the delay in response.

I don't see anything wrong this this in theory as ASCII art is pretty fun and the stakes in terms of backwards compatibility are low, but it's something I'll need to test to see if there are any edge cases that need resolving.

Et0h:
Tested for a bit and it all seems to be fine so I'll accept this improvement.

This change doesn't fix MOTD for older clients but there seems to be no reliable way of doing that. I can make it slightly better using Unicode hackery but that could cause problems for other implementations of Syncplay and still does not produce the same results without a unicode font face, so people with older clients will just get an uglier MOTD if you use ASCII - just call it abstract art!

PR #715, opened 2025-03-22
Fix inconsistent usage of tabs and spaces

Assistant:
Replaced tabs with 4 spaces in all the files where it makes sense (i.e., not in make files).
No other formatting changes were made, except a file was terminated with a newline that was missing it, which I couldn't avoid since my editor treats newlines as line terminators not separators.


Et0h:
Cheers - looks good to me.

Issue #714, opened 2025-03-22
Syncplay doesn't work with Memento (mpv-based) player


eyalmazuz:
Memento is a MPV-based video player with extra capabilities for language learning

https://github.com/ripose-jp/Memento

When trying to run syncplay with memento I get the following error

![Image](https://github.com/user-attachments/assets/25281af5-2128-4073-b2e0-b7d81657cdcc)

It would be great if you could support more players or alternatively, support players that are based on libmpv and not just straight up MPV itself.

Et0h:
Anyone can add support for a compatible mpv-based media player using the guide at https://github.com/Syncplay/syncplay/wiki/Guide-to-adding-support-for-a-new-media-player-to-Syncplay#supporting-a-variant-of-a-currently-supported-player

eyalmazuz:
> That player appears to have some pretty unique educational features so I can see why you would want to use it. I would be happy to provide advice to anyone wanting to add support for this player to Syncplay.
> 
> The starting point for anyone wanting to add support for a compatible.mpv-based media player is the guide at https://github.com/Syncplay/syncplay/wiki/Guide-to-adding-support-for-a-new-media-player-to-Syncplay#supporting-a-variant-of-a-currently-supported-player
> 
> Sometimes it just works and sometimes you need either workarounds to accommodate implementation-specific quirks and/or upstream changes in the media player itself so that it can be controlled via Syncplay.


Hi,
Thank you for the link, I've attempted to implement memento support copying the contents of mpvnet

but when running syncplay with the debug mode I get the following error:

```
eyal@eyal-xps157590 ~/D/P/syncplay (master)> python3 syncplayClient.py -d
[12時09分51秒] name 'pem' is not defined
[12時09分51秒] Not playing current index - Index none or length issue
[12時09分51秒] Failed to find index None in playlist
[12時09分53秒] player >> ['set_property', 'pause', 'yes']
[12時09分53秒] client/server >> {"Hello": {"username": "Eyal", "room": {"name": "HORSES"}, "version": "1.2.255", "realversion": "1.7.4", "features": {"sharedPlaylists": true, "chat": true, "uiMode": "GUI", "featureList": true, "readiness": true, "managedRooms": true, "persistentRooms": true, "setOthersReadiness": true}}}
[12時09分53秒] <mpv> Throttling message send, so sleeping for 0.05
[12時09分53秒] player >> ['script-message-to', 'syncplayintf', 'get_paused_and_position']
[...]

PR #716, opened 2025-03-24
Added support for memento player

eyalmazuz:
Adding support to [memento](https://github.com/ripose-jp/Memento) (mpv-based) player for syncplay

Solving #714


Et0h:
Nice work. I'll review this more when I have time but from a quick look I don't see any issues.

Et0h:
Okay, I just checked and the Windows path is: `C:\Program Files\Memento\memento.exe`.

It seems to work fine on Windows so far, but I'll give it a but more testing to confirm.

PR #734, opened 2025-06-23
mpv: Add support for clipboard property and Wayland

BlankEclair:
Pasting currently does not work when on Wayland, which this commit fixes.

mpv 0.40.0 also added support for the clipboard property, so I added support for that as well. I added fallbacks to the CLI versions, but I'm not too interested in getting a copy of mpv 0.23.0 to test it out (especially since the build system seems to rely on Python 2).


powerjungle:
I see, well I guess in my case it uses `xclip` (I have it installed), which in turn uses `Xwayland` and probably that's why it works for me now.

It's a nice change, I'll try to test it better when I can.

powerjungle:
Ah you mean the integrated chat in mpv, described here: [Enable chat input via mpv](https://syncplay.pl/guide/client/). Now it makes sense, sorry. 😅 Yes right now pasting doesn't work for me there. I'll try your version when I can. Thank you for the fix!

PR #743, opened 2025-10-22
Instead of localhost, use 127.0.0.1 (Fixes #302)

### LINKED ISSUE #302: Collaboration Syncplay 1.6.4a - VLC 3.0.8 on MAC shows problem
I'm new in this forum, so be kind and help me solve my problem 🙏
To synchronise external sound for Zoom-Meeting I wanted to use Syncplay in combination of VLC, both latest versions  1.6.4a resp. 3.0.8, on my MacBook Pro Mid 15 (MAC OS 10.14.6) and at friends with Mac and Win PCs.
I use VLC for long time and use actual updates. Syncplay was downloaded from Syncplay.pl homepage and installed as recommended by drag-drop the button on the programs button.
Trying to run Syncplay  opened the settings windows and asked for the file source.
When giving a local mp4-file and starting Syncplay connects to the desired server, starts the VLC app and the file is processed, but after short time two error messages appear (see screenshots) and no commands are taken any more.
In the documentation I found that Syncplay works with VLC 2.X. 
Can anyone give me a hint how to run Syncplay 1.6.4a and VLC 3.0.8 for synchronisation over internet ?

Regards  Wolfgang

Screenshot 2020-04-22 14 35 37
Screenshot 2020-04-22 14 36 10

smiba:
In issue #302 on MacOS (and presumably Linux), a wrongly configured hosts file can result in localhost not resolving

MacOS will for the most part function fine without it, allowing the system to otherwise function acceptably, while syncplay fails to work using VLC.

This change is slightly questionable though, as it could be said not having a definition of localhost is a failure on the operator of the system's part. 

Feel free to reject this PR in that case.

### DIFF
--- a/syncplay/players/vlc.py
+++ b/syncplay/players/vlc.py
@@ -528,7 +528,7 @@ class VlcPlayer(BasePlayer):
                 vlcoutputthread.start()
             self.__playerController._vlcready.clear()
             self._factory = VLCClientFactory(self.__playerController, self.vlcHasResponded, self.timeVLCLaunched, self.__process)
-            self.reactor.connectTCP('localhost', self.__playerController.vlcport, self._factory)
+            self.reactor.connectTCP('127.0.0.1', self.__playerController.vlcport, self._factory)
 
         def _shouldListenForSTDOUT(self):
             return not isWindows()
--- a/syncplay/resources/lua/intf/syncplay.lua
+++ b/syncplay/resources/lua/intf/syncplay.lua
@@ -90,7 +90,7 @@ local durationdelay = 500000 -- Pause for get_duration command etc for increased
 local loopsleepduration = 2500 -- Pause for every event loop (uses microseconds)
 local quitcheckfrequency = 20 -- Check whether VLC has closed every X loops
 
-local host = "localhost"
+local host = "127.0.0.1"
 local port
 
 local titlemultiplier = 604800 -- One week


Et0h:
Thanks for this suggested PR.

If this fixes a bug and doesn't create any new bugs then I'm happy for it to be accepted even if it is only to resolve a problem caused by the operator not properly configuring localhost. Fewer problems means less support requests, so everyone is happy.

I'm not aware of here aren't any setups which don't allow 127.0.0.1 for loopback and instead require ::1 IPv6 for loopback, but it's not something I've really researched in any detail.

smiba:
In my testing on MacOS it did fix the issue, when localhost does not resolve to anything.

>I'm not aware of here aren't any setups which don't allow 127.0.0.1 for loopback and instead require ::1 IPv6 for loopback, but it's not something I've really researched in any detail.

I guess the question might be what people are more likely to encounter, a IPv6 only loopback, or localhost failing to resolve. 

I would argue that not many systems will not have a IPv4 loopback, I've certainly never come across it... So this should be pretty safe, but at the same time no guarantees 😅

Et0h:
I guess we can always just cross that bridge if someone reports a problem with 127.0.0.1 - hard to plan for such a hypothetical without an actual real world example, so instead it makes sense to prioritise a confirmed if relatively obscure problem which has a simple fix.

PR #746, opened 2025-11-28
Add watched subfolders and watched history

Et0h:
## Watched subfolders and watched history

This feature makes it easier to keep track of videos that have already been watched, including by optionally storing them in a subfolder such as `Watched`.

It is effectively disabled by default, so existing behaviour is unchanged unless the relevant options are enabled.

Syncplay can keep a local history of watched filenames. Watched files are shown with a grey background in the playlist, and files can be marked as watched or unwatched by right clicking them.

The history is based on filenames, in line with the way Syncplay playlists identify files. Full original paths are not stored in the watched history.

In Settings -> Configuration you can choose the name of the watched subfolder, tell Syncplay to move watched files into it, and optionally allow Syncplay to create the subfolder when needed.

If automatic creation is disabled, you can just create watched subfolders in the directories where you want to use the feature. Syncplay searches media directories recursively, so it can still find files after they have been moved into the watched subfolder.

When watched folder moving is enabled, manually marking a file as watched or unwatched moves it to or from the watched folder. Manual moves happen immediately. If one fails, Syncplay reports the error.

Automatic moves work differently. If a file cannot be moved yet, for example because it is still locked by the media player, the move stays queued and Syncplay tries again later. A notification is shown once the move succeeds.

### Watched detection

A file is marked watched immediately when the local player reaches EOF through Syncplay's normal EOF handling.

There is also some additional checking for cases where a client may not get its own EOF event. One example is when another user reaches EOF first and pauses the room while this client is still slightly behind.

For that purpose Syncplay records actual playback near the end of the file. Time spent paused or elsewhere in the file does not count, but separate periods of playback near the end can add together.

For normal length files the near EOF window is the final 270 seconds. For shorter files it is limited to the final half of the file. The previous file check requires 120 seconds of accumulated playback in that area.

The existing check for changing playlist item while within five seconds of the end is also kept. It now requires the file to have been loaded for at least 120 seconds and requires at least five seconds of actual playback near the end. This avoids treating a file as watched just because it was opened or seeked to the end and left paused.

Normal EOF and playlist advancement behaviour are otherwise unchanged.

### Removing watched items from the playlist

There is an optional setting to remove a watched item from the shared playlist after EOF. This is disabled by default.

When enabled, Syncplay waits until the expected next playlist item has been selected, the room has settled on it, and the client has received confirmation of its own playlist index change before removing the watched item.

If that does not happen within the timeout, the item is left in the playlist. This is intended to avoid interfering with the normal EOF transition or removing an item while clients are still on it.

The watched features can therefore be used just for local history, for watched folder handling, or together with automatic playlist cleanup.


Et0h:
@Namyts I've now added an "Automatically remove watched files from playlist" option to this branch (something you asked for in #767). It needs to be enabled in the 'Files' config tab.

SquirrelKiev:
@Et0h Small thing but doesn't seem that the mark as watched option in the playlist right click menu works with multiple files selected? seems to only mark the first selected one as selected.
image

Et0h:
Thanks for letting me know @Namyts, it'd slipped my mind to add support for marking multiple files but in hindsight it's going to be a common use case for those just starting to use the feature. I'll look into correcting this oversight.

Et0h:
@SquirrelKiev Okay, you should now be able to mark multiple files as watched/unwatched without issue. As we've merged future issues should be raised as new issues.

Issue #406, opened 2021-03-26
[Feature Request] Ability to sync video speed


jvschiavo:
Currently there is no possible way to sync video speed across different clients.

The only way to watch videos together in 2x for example would be to everyone manually changing their video player speed to the desired rate or the constant cuts and jumps trying to sync would make it impossible to watch.

But if someone likes to change the video speed frequently depending on the part of the video (like me), the best solution would be to sync the video speed across clients.

Would something like this be possible to implement?

Thanks again for the great software. :)

Et0h:
It's possible, but it isn't easy and would need a lot of testing to ensure it works across all media players, is compatible with managed rooms, changing rooms mid-playback, joining mid-playback (when first loading the file or following a disconnect when the file is already open), and how to handle slowdowns. You would also need a system to determine if some people in the room have an older version of Syncplay to warn them that they will be out of sync. Whoever takes on this project would need to be willing to maintain the feature in the future, which hopefully won't be as difficult as implementing it in the first place.

https://syncplay.pl/about/ideas/ lists "Allowing rooms to have a custom playback rate" as something that "Would need someone to come on board to develop". To date, nobody has volunteered so it isn't happening.

See https://github.com/Syncplay/syncplay/issues/339 for the previous discussion of this suggestion. I have now tagged #339 as 'help wanted' which means if someone is looking for something to volunteer on then they will see it as an option. I'll leave this issue open for a time to see if it sparks interest in anyone wanting to put their time into this project, but if nobody steps forwards then I will close this issue.

Issue #339, opened 2020-07-23
[suggestion] Possibility to increase playback speed


alekskomp:
Hi
Can you please add possibility to increase playback speed? 
x1.25 x1.5 x1.75 etc.
Or add syncing speed changes

daniel-123:
Adding "proper" support for playback speed synchronization is considerably large undertaking in current code base.

That said - if you disable the "Slow down on minor desync" option, then Syncplay will not mess with playback speed of your player. If everybody in the room sets the same playback speed manually, then syncing should work just as if at playback rate of 1.

Et0h:
I can see why people want this feature, but I agree with @daniel-123 that it would take quite a lot of work to implement and test as a lot of Syncplay's code is based on an assumption of playback being 1 second per second. This is one of those things that could end up inadvertently creating lots of bugs and so would need a lot of testing.

Et0h:
As noted at #406:
> It's possible, but it isn't easy and would need a lot of testing to ensure it works across all media players, is compatible with managed rooms, changing rooms mid-playback, joining mid-playback (when first loading the file or following a disconnect when the file is already open), and how to handle slowdowns. You would also need a system to determine if some people in the room have an older version of Syncplay to warn them that they will be out of sync. Whoever takes on this project would need to be willing to maintain the feature in the future, which hopefully won't be as difficult as implementing it in the first place.
> 
> https://syncplay.pl/about/ideas/ lists "Allowing rooms to have a custom playback rate" as something that "Would need someone to come on board to develop". To date, nobody has volunteered so it isn't happening.

PR #754, opened 2026-02-14
Sync playback speed across room

### LINKED ISSUE #115: [Feature request] custom playback speed
Usually when my group of friends watch movies we watch at 1.1-1.2X speed. This feature would be very desirable for us for syncplay to support. It seems that if everyone sets the speed at the same time it keeps us sync until it seeking or pausing resets it; but have a synced setting would be best.

hex110:
Fixes #115 , #339 , #406 

## Summary

  - Adds playback speed as a synced room property, similar to pause and seek
  - When a user changes speed in their player, it propagates to all clients in the room
  - In controlled rooms, only operators can change speed
  - Slowdown-on-desync works relative to the room speed (e.g. room at 1.2x, slowdown = 1.2 * 0.95)
  - Dead-reckoning on both client and server uses speed for position extrapolation
  - Backward compatible: missing speed field defaults to 1.0 — old clients/servers work fine

## Player Support
*   ✅ **Supported:** mpv, mpv.net, IINA, Memento, VLC, MPlayer
*   ❌ **Unsupported:** MPC-HC, MPC-BE (limitation: `speedSupported = False` in current API implementation)

## Files Changed
*   `server.py`: Room/ControlledRoom store speed; Watcher propagates speed changes.
*   `protocols.py`: Added Speed field in State message (client + server).
*   `client.py`: Global speed state, dead-reckoning, slowdown relative to room speed.
*   `players/*.py`: Added thread-safe speed detection via properties/events for mpv, VLC, and MPlayer.
*   `resources/lua/*`: Updated Lua scripts to report speed/rate in status responses.
*   `constants.py`: `DEFAULT_PLAYBACK_SPEED = 1.0`
*   `messages_en.py`: Added `speed-change-notification` message.

## Test plan
I have verified the following scenarios:

- [x] **Basic Sync:** Two mpv clients on local server — speed change on one syncs to the other.
- [x] **Permissions:** Controlled room — non-operator speed change is rejected.
- [x] **Drift/Lag:** Slowdown-on-desync calculates correctly relative to room speed.
- [x] **Compatibility:** Old client (no speed field) connects and defaults to 1.0 without errors.
- [x] **Unit Tests:** Added `tests/test_speed_sync.py` and all tests pass.

Didn't verify VLC or MPlayer as none of my friends use it and I'm not familiar with how they work at all, so I'd appreciate it if someone can verify these players as well.


Et0h:
Hi, it will take me some time to go through all this, but thanks for helping to progress this long-requested feature.

Based in part on my experience from shared playlists my initial thoughts in terms of checking to ensure things work as expected are (in no particular order):
* Reliability on disconnect/reconnect (including room being temporarily empty and therefore deleted)
* Reliability on joining and changing rooms to those with same/different speeds
* Reliability when changing from one media file to another
* Backwards/forwards compatibility with older/newer clients and servers (inc warning messages using chat as appropriate), inc. what happens for people using incompatible clients, potentially making use of the 'features' system
* Whether speedSupported can be set to true for MPC-BE
* Control via GUI and CLI
* Compatibility with (mostly deprecated) offset feature
* Whether the speed variable needs to be explicitly typed to a float in setSpeed
* Ensuring speed-up/slowdown code does not improperly trigger 'speedChanged = speed is not None and speed != self._room.getSpeed()'
* Ensuring no .7z binaries in the PR

gavtroy:
The latest commit is missing a couple of lines resulting in:

> 'UiManager' object has no attribute 'updateSpeed'

Also while changing speed from inside mpv works well, using the syncplay GUI or "/speed" doesn't stick. The peers see "userA changed playback speed to 1.1x" immediately followed by "userA changed playback speed to 1.0x". I think the speed change never gets sent to our own player?

Otherwise it seems to be working well.


By the way you might consider storing the room speed in the database so it can be restored across restarts. It's not super necessary imo but probably fairly easy to do. Edit: [like this?]( https://gist.github.com/gavtroy/a7c90456fe4c6169750d9fa4a59c1d1c)

PR #756, opened 2026-02-16
Add /qan and /qans commands for queue-add-next

### LINKED ISSUE #702: Add new entry next in the queue with /qan
I often find myself adding a video to the queue with /qa, and then dragging the video up to the point in the queue below the current video. A /qan command would simplify the process. 
A /qans command (queue add next and select) would also be useful, though the command would probably need a better name.

hex110:
## Summary
- Adds `/qan [file/url]` — inserts a file/URL after the currently playing item in the playlist                                                                                                                                                                                         - Adds `/qans [file/url]` — same as `/qan` but also selects the new item for playback                                                                                                                                                                                                
- If nothing is playing, the item is inserted at the beginning of the playlist

Follows existing naming conventions (`/qa`, `/qas`, `/qn`).

Closes #702


Et0h:
Hi @hex110 I finally had time to test it in more detail (sorry for the delay) and I noticed two issues that need to be resolved:
* Running /qans on its own adds None to the playlist.
* Running /qans selects the last playlist item, not the item just added (unless the item is already in the playlist in which case it switches to it)

PR #759, opened 2026-02-16
Use fakeroot when building .deb packages

### LINKED ISSUE #513: Debian packages are not created with root privileges
**Description**
The debian packages aren't built with root:root privileges. So by installing them with dpkg (or equivalent) the files are extracted and the UID and GID are set to 1001:123, which can be literally anything on the system.

**To Reproduce**
* download any .deb installer package
* install (e.g. dpkg -i syncplay_1.6.9.deb)
* check file permissions with ls -l /usr/lib/syncplay or ls -l /usr/lib/syncplay-server

**Expected behavior**
Root ownership of the extracted files

**Version and platform:**
 - OS: Pop!_OS 21.10
 - Syncplay version and build type: 1.6.9

**Additional context**
The files seem to be created with runner:docker privileges.

**Solution recommendation**
The .deb packages could be built with fakeroot, where it is possible to easily set UID and GID to 0:0, which is always the systems root, and build the .deb package.
I've extracted and rebuilt the package with correct permissions and installed it successfully on my system. It's working without issues.

hex110:
## Summary
- Without `fakeroot`, files in the `.deb` package are created with the build user's UID:GID instead of `root:root` (0:0), causing incorrect ownership when installed via `dpkg`
- Wraps both `make install` and `dpkg -b` with `fakeroot` in the client and server deb build scripts

Closes #513


Et0h:
I'm not a Linux user, but would it be better to use `dpkg-deb --root-owner-group -b` instead of `fakeroot` here to avoid an additional build dependency?

Under this alternative, the only changes from `master` would be to the final package-building command in each Debian packaging script:

```diff
# ci/deb-script.sh
 make install DESTDIR=/tmp/syncplay
-dpkg -b /tmp/syncplay/
+dpkg-deb --root-owner-group -b /tmp/syncplay/

# ci/deb-server-script.sh
 make install-server DESTDIR=/tmp/syncplay-server
-dpkg -b /tmp/syncplay-server/
+dpkg-deb --root-owner-group -b /tmp/syncplay-server/
```

The `make install` / `make install-server` staging steps and the generated `DEBIAN/prerm` scripts would remain unchanged.

For clarity, this sets the ownership of filesystem entries recorded in the package to `root:root` (i.e. it does not cause Syncplay or the media player processes it launches to run as root).

Debian documents `--root-owner-group` as:

> Set the owner and group for each entry in the filesystem tree data to root with id 0

It also notes that the option is useful for rootless builds: https://manpages.debian.org/trixie/dpkg/dpkg-deb.1.en.html

This directly addresses #513 by preventing the CI user’s UID/GID from being stored in the package.

The option requires dpkg 1.19.0 or newer, which should be fine for current Debian and Ubuntu CI environments. Since Syncplay does not appear to ship files that intentionally require non-root ownership, this seems cleaner and more focused than adding `fakeroot`.

PR #761, opened 2026-02-16
Add Romanian translation

hex110:
## Summary
- Adds complete Romanian (Română) language support with all ~457 UI strings translated
- Registered as language code `ro` in the message system
- Follows the same pattern as existing translations

Done using Claude Code and verified manually by myself and a friend, making sure everything sounds like we'd expect it to and that nothing sounds weird/misused.

### DIFF
--- a/syncplay/messages.py
+++ b/syncplay/messages.py
@@ -11,6 +11,7 @@ from . import messages_it
 from . import messages_pt_PT
 from . import messages_pt_BR
 from . import messages_tr
+from . import messages_ro
 from . import messages_ru
 from . import messages_zh_CN
 from . import messages_ko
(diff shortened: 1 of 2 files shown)


Et0h:
@hex110 I had a look over it (sorry for the delay) and here's the outstanding things to do:
* Update docs/syncplay.1 so the --language valid-values list includes ro, as required by the translation guide at https://github.com/Syncplay/syncplay/wiki/Guide-to-translating-Syncplay-into-another-language
* ChatGPT tells me there is some inconsistent register between informal tu/ești/tău and formal dumneavoastră/sunteți/vă. Is this criticism valid?

**The review stated:**

The main quality issue is inconsistent register: it alternates between informal tu/ești/tău/ta and formal dumneavoastră/sunteți/vă. For example, readiness messages use informal ești, while file mismatch and room messages use formal dumneavoastră / Sunteți. Similar mixing appears in settings/tooltips, with informal te/tău near more formal UI phrasing.

That is not a functional problem, but it makes the translation feel AI-assisted.

A few wording notes:

- operator al camerei for “controller” is understandable, but decide whether Syncplay wants “operator”, “moderator”, or another term and use it consistently.
- Sari la timp / salt is understandable, but Derulează/Sari terminology should be checked for consistency with Romanian media-player conventions.
- creați un issue is understandable developer language, but raportați o problemă may be more natural for general users.
- stream and URL are probably acceptable as technical terms.

PR #771, opened 2026-04-21
Move database room deletions to be consistent with saves

gavtroy:
Currently emptying a room with an empty playlist fires off two transactions which get run together:
- update the room db entry (with nothing useful)
- delete the room db entry

Usually the DELETE is processed first and then INSERT OR REPLACE adds the room back to the database. Although the room is gone from the GUI, it will pop back into existence when the server is restarted.

This change simplifies the logic to always delete the room from the database when the playlist is empty (regardless of whether there are watchers).

### DIFF
--- a/syncplay/server.py
+++ b/syncplay/server.py
@@ -489,20 +489,16 @@ class RoomManager(object):
             if RoomPasswordProvider.isControlledRoom(roomName):
                 room = ControlledRoom(roomName, self._roomsDbHandle)
             else:
-                if roomName in self._rooms:
-                    self._deleteRoomIfEmpty(self._rooms[roomName])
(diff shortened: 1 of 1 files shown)


Et0h:
Nice catch!

Et0h:
Is this compatible with https://github.com/Syncplay/syncplay/pull/754 which adds playback speed support?

gavtroy:
Yes it applies and works fine. That PR doesn't bother storing the room speed in the database, but even if that gets added it shouldn't be an issue.

Et0h:
@gavtroy Great, I've tested it and it seems fine so I'll be accepting the PR. Thanks for your work on this!

PR #772, opened 2026-04-23
Inline gif (media) rendering, unique chat colors for each user

Osanosa:
image

### DIFF
--- a/syncplay/ui/gui.py
+++ b/syncplay/ui/gui.py
@@ -1,3 +1,4 @@
+import hashlib
 import os
 import re
 import sys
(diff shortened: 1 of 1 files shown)


Et0h:
It's neat that you can do this but I consider adding media embedding to be unnecessarily risky feature creep.

In my view the change goes against key principles at https://github.com/Syncplay/syncplay/issues/315  because:

1. Text only chat works with the gui, command line version of Syncplay and with the mpv osd, but images/GIFs would show up differently depending on the chat display method. This goes against the principle of consistency/predictability.
2. This feature is not very backwards compatible with previous versions of Syncplay. This goes against the principle of compatibility.
3. Allowing media adds a whole new security and mischief vector we just don't need to have to find mitigations for. This goes against the principles of disruption minimisation and potentially privacy.
4. If people want Discord they can use Discord, there is no need for this in Syncplay. This goes against the principle of simplicity.

Osanosa:
Like idk it just parces direct file links and doesn't break anything 
I see no issues
On mobile client I did also add a picker and I'm loving it! (https://t.me/ThemedProject/1019)

So I don't see any compatibility issues as where not supported it is shown as text and where is supported it is a fun thing 

Do you find this problematic still?
I mean, nobody's forcing users to send those 🤪

Osanosa:
@Et0h any more feedback? it is fully compatible with previous versions, as it just detects direct media URLs.

daniel-123:
@Osanosa I'm going to back up Et0h with comment that this is a security nightmare. Downloading and rendering media is absolutely rife with vulnerabilities.

To me, all other arguments around this are just fluff - inevitable security issues alone make this completely unsustainable burden to maintain for current core Syncplay developers.

Osanosa:
Okay this could be an optional toggle that users take risks to use it?

Osanosa:
but it loads only media types? like cuh?

Et0h:
Sorry but I really don't want to compromise on security.

PR #778, opened 2026-05-26
Initialize the variable latencyCalculation before the conditional

Kcchouette:


Et0h:
I need to test it, but looking at the code this seems like a good bug fix.

PR #779, opened 2026-05-26
Sanitize message and file path inputs

Kcchouette:
Remove carriage return and newline characters from messages and file paths.


daniel-123:
Does this change solve a specific problem you encountered?

In messages I could see newlines as potentially useful.

In filenames, *technically* both `\r` and `\n` are valid characters. Normalizing filenames is an awfully complex issue that's OS and filesystem dependent and I would prefer not to open that specific can of worms without a very good reason.

Kcchouette:
@daniel-123 no, this is only a sanitize fix

PR #781, opened 2026-06-12
Speed sync

sofoxe1:
branches of: https://github.com/Syncplay/syncplay/pull/754 (3fddded2f185ce9878fa73848c8121088fdffcb8)
fixed: tests failing to run on linux, speed change not being showed in chat of the user who changed it, 'UiManager' object has no attribute 'updateSpeed', ui failing to set the speed

i'm a surprised that pr was abandoned as it is almost done i only had to change few lines

i don't really know python or this project so no idea how good the code is but it seems to work (only tested with mpv)


sofoxe1:
i have been using it for a while and it works without any issues

PR #783, opened 2026-06-20
Fix/Update ES localization

Readek:
Like I've been using this software for years now and I'm just here only to fix "resumido" when someone unpaused as that's literally not a word and it was driving me mad but I guess I also ended up translating all missing texts I could find.

I would do a way more intense review but it seems like the original translation leans more into Latinamerican Spanish, and it wouldn't be right for me to change stuff, me being from Spain. Unless this translation could split into `es-la` and `es-es`.


Et0h:
@Readek I have nothing against the translation being split into es-la and es-es. Is that something you feel able to do or do you need me to do it in a branch?

Readek:
actually i have no idea how would i code and implement a new language, i have never coded with Python

Readek:
Hey, is this PR going anywhere? Just merging it would be enough, no need to make things complicated by adding an entire new language option.

Et0h:
Hi @Readek - The recently released 1.7.6 was a bugfix only, and we're now working on a 64-bit installer for 1.7.6 and after that we can start merging stuff into 1.7.7 development version to give it plenty of time to be tested before it is released.

Et0h:
@Readek I've now created `messages_es_ES.py` for you to modify in your fork and make a new PR. You will need to update your branch:
1. Go to your Syncplay fork.
2. Make sure they’re on the branch they want to update, usually master.
3. Click Sync fork above the file list.
4. Click Update branch.

Issue #585, opened 2023-02-04
On MacOS Ventura; Error: "Failed to recognize file format" when attempting to play any YouTube video


PuffyRainbowCloud:
**Describe the bug**
When attempts are made to play a YouTube video on either of my machines running macOS Ventura using MPV I receive the error message "Failed to recognize file format" and nothing plays. I can play the same videos fine using SyncPlay and MPV on Windows 10. This applies to SyncPlay version 1.6.8, 1.6.9, and 1.7 Beta 1.

**To Reproduce**
1. Launch SyncPlay.
2. Fill out relevant server- and room information.
3. Choose MPV as the video player.
4. Attempt to play a YouTube URL via any of the methods within the UI of MPV or SyncPlay.

**Expected behavior**
The linked YouTube video should start playing.

**Version and platform:**
 - OS: macOS Ventura 13.0
 - Syncplay version and build type: Syncplay 1.6.8
 - Media player and version:  MPV 0.35.0

Et0h:
Does it work fine if you're running mpv on its own without Syncplay? Do you have ytdl installed?

girlmaya:
> Does it work fine if you're running mpv on its own without Syncplay? 

Yes. It doesnt even reach mpv as far as i can tell. Can provide info just lmk i would really like this fixed

ReeverM:
I'm having the exact same issue on Windows 10. When I /qa it, it removes it from the queue immediately seemingly without reaching mpv (just the "x has updated the playlist" message pops up).

When I turn off shared playlists, /qa the youtube video and then turn on shared playlists again, it DOES reach mpv but returns "failed to recognize file format" but in this case does NOT remove the entry from the playlist.

I initially thought it was related to ffmpeg, so I installed that and now mpv plays pasted youtube videos fine (lua script to paste) but the playlist issue prevails.

girlmaya:
> on Windows 10. 

> it DOES reach mpv but returns "failed to recognize file format" 

I think this is a different bug to what i'm experiencing on macOS. Not sure about the original issue creator

suhailskhan:
Having the same issue with macOS Sonoma. My mpv is able to play from YouTube URLs without Syncplay.

- macOS Sonoma 14.1
- Syncplay 1.7.1 RC1
- mpv 0.36.0
- yt-dlp 2023.10.13

girlmaya:
Any update on this?

suhailskhan:
[...]

PR #791, opened 2026-08-02
macOS: Fix yt-dlp launch through mpv

Et0h:
This is a fix inspired by #785 to address #585 and is hopefully more reliable.

This PR removes deprecated handling for youtube-dl on macOS and updates environment variable management for mpv subprocesses.

### DIFF
--- a/syncplay/players/mpv.py
+++ b/syncplay/players/mpv.py
@@ -6,7 +6,6 @@ import sys
 import time
 import subprocess
 import threading
-import ast
 
 from syncplay import constants
 from syncplay.messages import getMessage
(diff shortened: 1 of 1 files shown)


syncplay/players/mpv.py
```diff
@@ -601,30 +600,25 @@ def __init__(self, playerController, playerIPCHandler, playerPath, filePath, arg
...
-                    pythonPath = None
-                if pythonPath is not None:
-                    env['PATH'] = python_executable + ':' + env['PATH']
-                    env['PYTHONPATH'] = pythonPath
+                # Remove py2app's bundled Python environment from mpv subprocesses.
+                env.pop('PYTHONHOME', None)
+                env.pop('PYTHONPATH', None)
+            
+                # Add common executable locations while preserving inherited PATH entries.
+                macPaths = [
```
prez, line 609: why is this necessary?

Et0h: I was trying to address the macOS yt-dlp/mpv problem discussed in #585 and #785, while making the fix work more universally rather than depending on a particular yt-dlp/Python installation.

For context, I'm doing my best to get something workable while not being a macOS user let alone developer.

prez: I know what you're trying to accomplish, I authored that PR :) 
my question is about the addition in this PR relative to mine (extending PATH with 'common' paths) - what makes you think PATH is not sufficiently populated therefore needs to be extended?

Et0h: Fair question. The PATH change isn't needed for the "PYTHONHOME"/"PYTHONPATH" fix itself. I added it because removing the old workaround also removes its explicit "/opt/homebrew/bin:/usr/local/bin:/usr/bin" PATH handling.

My concern is the normal DMG/Finder-launched use case: py2app documents that applications launched normally get a minimal environment and don't pick up changes from the user's shell profile. This was also discussed in #276, where it was noted that macOS GUI apps may not have "/usr/local/bin" in PATH and the suggestion was to append it rather than replace the existing PATH.

So the intention was to preserve the old code's ability to find a Homebrew-installed downloader, while avoiding overwriting any PATH that is already present. "/opt/homebrew/bin" covers current Apple Silicon Homebrew and "/usr/local/bin" covers Intel Homebrew.

On reflection, I think the patch can be narrower here: rather than adding all of the standard system paths, it could just preserve the inherited PATH and append "/opt/homebrew/bin" and "/usr/local/bin" if missing.

What's your thoughts on it being tweaked to:
if isMacOS():
    env.pop('PYTHONHOME', None)
    env.pop('PYTHONPATH', None)

    # Add common Homebrew locations while preserving inherited PATH entries.
    pathEntries = [path for path in env.get('PATH', '').split(os.pathsep) if path]
    for path in ['/opt/homebrew/bin', '/usr/local/bin']:
        if path not in pathEntries:
            pathEntries.append(path)
    env['PATH'] = os.pathsep.join(pathEntries)

prez: macos applications, launched from dock, finder, or as login items generally don't pick up the PATH from the user's shell profile as you have noted. 
this is intended by apple and not specific to py2app. the proper mechanism to configure this is via `/etc/paths` and `/etc/paths.d` (see `man path_helper`).
if homebrew users have not added homebrew to the system-wide path, then that is indicative of a misconfigured system, in my opinion. now, whether this is common or not, I wouldn't be able to tell you - I do not use homebrew (and am most knowledgeable about Linux too). and I'm not installing homebrew just to try this, but it's worth noting that some package managers (and proprietary software packages) install their paths into `/etc/paths.d` automatically.
should syncplay accommodate for this misconfiguration of unknown commonness? probably not. but it's not a hill I'm willing to die on. so, your call :)

Issue #611, opened 2023-06-09
Client UI breaks when moving between monitors with different resolution.


Azi9Intent:
**Describe the bug**
The Client UI becomes nearly unusable when I move the window from my 1920x1080 monitor to my 3440x1440 monitor. It also refuses to open on my main monitor, choosing to only open on the secondary one (the 1080p one). 

**To Reproduce**
(1.?) Have multiple monitors with differing resolutions?
2. Open client on 1080p monitor
3. Drag client to other monitor

Unsure if reproducible going from higher res to lower, as I cant seem to have it start on the other monitor.

**Expected behavior**
The UI doesn't break

**Screenshots**
![syncplayui](https://github.com/Syncplay/syncplay/assets/34526091/a15e9583-e4e7-41ee-9bf2-c5afc8218b3c)

Screenshot is of the full window after moving between monitors

**Version and platform:**
 - OS: Windows 10 64 bit
 - Syncplay version and build type: Syncplay 1.7.0 Portable
 - Media player and version:  N/A

**Additional context**
N/A

Et0h:
I only have one monitor so I'm not in a good position to fix this issue. Hopefully someone wish two monitors will figure out a way of fixing it without stuff.

Et0h:
This should be fixed by #795.

PR #795, opened 2026-08-11
Move Windows build to 64-bit PySide6 to resolve high-DPI/scaling issue (fixes #793 by SquirrelKiev and #707 by ritiek)

### LINKED ISSUE #793: Syncplay doesn't handle non-100% scaling very well
**Describe the bug**
On certain display configurations (tested with main monitor at 150% scaling and resolution at 3840x2160), syncplay scales incorrectly.

**Screenshots**

Syncplay on 150% scaling monitor
Image
Said window dragged to 2nd monitor

Image

Syncplay opened on 2nd monitor, 1920x1080, 100%

Image

**Version and platform:**
 - OS: Windows 11 25H2
 - Syncplay version and build type: Syncplay 1.7.6 (reproduced on 1.7.5 and 1.7.4 however)

**Additional context**
From some personal testing, this seems to have been fixed in PySide6 (Syncplay is PySide2). Been using a hacked together fork with a newer python version and PySide6 which I've been using for a while, has been working fine for me so far?

Image

Et0h:
Adds a new 64-bit Windows build using Python 3.11 and PySide6, resolving the Windows high-DPI/scaling issue reported in #793 by @SquirrelKiev. 

As part of this work I also resolves the same issue that @ritiek was trying to fix in #707 about the invocation of the font dialogue being broken in PySide6. The holding issue on that PR was that the change didn't work on PySide2.

The existing 32-bit Python 3.8 / PySide2 build is retained as a fallback for now, but might be removed in the future.

## Summary

- Add a parallel Windows x64 build using:
  - Python 3.11
  - PySide6
  - py2exe 0.14.0.0
- Keep the existing Windows x86 / PySide2 build unchanged.
- Update `buildPy2exe.py` to support packaging either PySide2 or PySide6.
- Package the Qt6 Windows platform and modern Windows style plugins.
- Use Qt6's native high-DPI handling rather than applying the old Qt5 high-DPI attributes.
- Fix PySide6 handling of persisted `QFont` weights (resolving the issue of #707 without the compatability issues).
- Fix Windows/Qt6 window flags for the main window and custom dialogs.
- Prune unused Qt6 libraries from the packaged build.
- Add a packaged x64 GUI startup smoke test to the Windows build workflow.
- Fix generated Windows uninstaller directory removal.
- For fresh x64 installations, default to the native 64-bit Program Files directory.
- When upgrading an existing installation, continue using its registered installation directory

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -87,7 +87,7 @@ jobs:
           pip3 install -U setuptools wheel pip
           pip3 install -r requirements.txt
           pip3 install "PySide6-Essentials>=6.11,<6.12"
-          pip3 install "py2exe==0.14.2.0"
+          pip3 install "py2exe==0.14.0.0"
 
       - name: Check Python dependencies
         run: |
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -95,6 +95,13 @@ jobs:
           python3 -c "from PySide6.QtCore import __version__; print(__version__)"
           python3 -c "from PySide6.QtCore import QLibraryInfo; print(QLibraryInfo.path(QLibraryInfo.LibrariesPath))"
           python3 -c "import py2exe; print(py2exe.__version__)"
+          
+      - name: Check PySide6 Windows plugins
+        shell: pwsh
+        run: |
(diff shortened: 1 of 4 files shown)


SquirrelKiev:
should fix #611 too

Issue #796, opened 2026-08-11
Folder search permanently disables after a transient scan timeout (requires manual re-affirmation)


AlySerry0:
### Summary

When a media-folder scan hits `FOLDER_SEARCH_FIRST_FILE_TIMEOUT` or `FOLDER_SEARCH_TIMEOUT`, `FileSwitchManager._updateInfoThread` sets `folderSearchEnabled = False`. That disables automatic file switching entirely until the user manually re-affirms their directories via **File → Set Media Directories → OK**. I'd like to propose making this **self-healing**: abort the slow pass, but keep the feature enabled so the next periodic pass can recover on its own.

### Why the current behavior hurts

Timeouts here are usually *transient*. A network share or a spun-down external drive is slow on the first **cold** scan, but the very next periodic pass typically finishes in a couple of seconds once the OS/drive metadata caches are warm. Under today's behavior the first cold pass **permanently** disables the feature, so that fast warm pass never runs — the user is left with "not found" playlist items and a menu dance to recover.

This is a long-standing pain point — see #130 (same scenario: a Samba mount with thousands of files, asking for a bigger timeout).

To be upfront about root cause: in my case the real bottleneck was **server-side** (an NTFS-over-USB drive via ntfs-3g on a Raspberry Pi — cold directory enumeration ~30–50s, warm ~1–5s), and I fixed that on the server. I'm **not** asking Syncplay to make slow drives fast. This is purely about the *client UX* when a transient timeout does occur, which can affect anyone with a large or networked media directory.

### Proposed behavior

- On timeout, `return` from the pass (as today) but **don't** set `folderSearchEnabled = False`.
- Surface the timeout notice only **once** (tracked with a `folderSearchTimedOut` flag) so it doesn't repeat every interval.
- **Clear** the flag whenever a pass completes within the timeout, so it keeps self-healing and will still notify again on a future, distinct episode.

This keeps the original intent — the slow pass is still aborted, and the once-only notice plus the existing interval cadence keeps I/O bounded — while removing the permanent-disable and the manual re-affirmation requirement.

Follow-up: the `folder-search-timeout-error` message currently tells users to re-affirm via the menu, which would no longer be necessary. I'm happy to reword it (across locales) as part of the change.

### Implementation

I have a small patch ready against `master` (client.py only, +17/−4):
https://github.com/Syncplay/syncplay/compare/master...AlySerry0:syncplay:folder-search-retry-not-disable

[...]

Issue #130, opened 2017-01-29
Increase FOLDER_SEARCH_TIMEOUT again


Ghosthree3:
I know that this was recently increased (which helped a _lot_) but is there any reason it can't be increased further? It seems to be 6 seconds now which works for me most of the time, but about 10% of the time I still get the folder-search-timeout-error. Is there some limitation that means it can't be even 60 seconds? I understand the 15s FOLDER_SEARCH_FIRST_FILE_TIMEOUT but not FOLDER_SEARCH_TIMEOUT being so low.

FYI I'm using a samba mount with 4852 files in it currently as my media directory. I can only assume the fact it's a network mount is the reason sometimes it works and sometimes it doesn't. Generally if I reopen syncplay after it errors it'll work fine.

Et0h:
If it gives an error then I think that you don't need to restart Syncplay to get it to work again, you just need to go File -> Set Media Directories, and press 'Okay'. That should reset the flag which told Syncplay to ignore the network drive, and things should work again.

Anyway, getting to your suggestion...

**Current default options for the File Switch / Folder Search feature are:**
- FOLDER_SEARCH_FIRST_FILE_TIMEOUT = 15.0 # Secs - How long to wait to find the first file in folder search (to take account of HDD spin up)
- FOLDER_SEARCH_TIMEOUT = 6.0 # Secs - How long to wait until searches in folder to update cache are aborted (after first file is found)
- FOLDER_SEARCH_DOUBLE_CHECK_INTERVAL = 30.0 # Secs - Frequency of updating cache

Generally speaking it is recommended that users use local drives rather than network shares as their media directory, and avoid directories with way too many files in them. Values could be increased, but if the value for FOLDER_SEARCH_TIMEOUT is greater or equal to FOLDER_SEARCH_DOUBLE_CHECK_INTERVAL then it could result in situations where Syncplay is constantly updating the cache, and I'm guessing that this isn't a good thing.

One potential way that Syncplay could address this would be to update the cache less frequently in the event of things taking too long, and/or potentially changing how frequently it refreshes the cache depending on if there are any unfound files. The reasons this doesn't already happen is because (1) it would result in less consistency of behaviour, (2) it could encourage people using folders which are too big, and (3) it is more complicated and therefore harder to code and test.

Ghosthree3:
[...]

PR #799, opened 2026-08-17
Media directory monitoring: notifications + adaptive reconciliation, no disable on timeout

AlySerry0:
Draft implementing the media-directory monitoring spec from the #796 discussion (refs #130). This revision replaces the earlier draft on this branch and is rebased onto current `master` (so it builds on the merged watched-subfolders `EpisodeFilenameParser`).

Developed with AI assistance (Claude) and reviewed/tested by me, following the updated agent docs (#763).

## Approach

Filesystem notifications are used as an **acceleration path** over authoritative reconciliation scanning; a full recursive scan stays the completeness authority, and absence of an event is never treated as proof nothing changed.

**Kept from the earlier draft:** transient timeouts no longer disable folder search, the synchronous single-flight scan claim, watchdog support, and the Windows packaging.

**Reworked per the spec:**

- **`syncplay/filemonitor.py` (new)** — a small, playlist-blind `FileMonitor` owning the watchdog observer lifecycle and the thread→reactor handoff. It knows only paths/watch characteristics: a recursive watch per media root, plus a budgeted set of supplemental non-recursive watches for Windows network roots (`FOLDER_SEARCH_NETWORK_WATCH_LIMIT = 32`). If watchdog is unavailable or a watch fails, it provides no events and scanning continues.
- **Event-driven cache** — clear file create/delete/move events update `mediaFilesCache` directly and idempotently on the reactor thread (no scan to re-confirm; component/case-aware path matching so `P:\TV` ≠ `P:\TV2`). Directory events use targeted subtree reconciliation, with full reconciliation as the ambiguity/failure fallback.
- **Concurrency** — single-flight full/targeted worker; events arriving during a scan collapse into one pending full reconciliation; workers only read the filesystem and return results, the reactor thread commits (no shared lock, since qt5reactor makes the GUI and reactor one thread). Downstream `fileSwitchFoundFiles()` is coalesced (~1s).
- **Adaptive polling** — full reconciliation runs at a slow safety interval (`FOLDER_SEARCH_RECONCILIATION_INTERVAL = 300`) while the current/next local files resolve, and returns to the existing ~30s `folderSearchDoubleCheckInterval` while a needed file is unresolved (using `findFilepath` resolution semantics; URLs don't force urgent polling).
- **Timeout/warning** — degraded operational state (cleared only after three consecutive clean full scans) is separated from a single user-visible warning per media configuration/session. `folderSearchWarningThreshold` becomes additional grace on a 3s base delay, so **no config migration**. The `folderSearchUseNotifications` option added in the earlier draft is removed (notifications are an internal optimisation). The warning no longer says folder searching has been disabled.
- **Watch prioritisation** — supplemental network-directory watches are ranked by current-file/current/next-playlist relevance and same-series/season directories via a small `getRelatedEpisodeDirectories` helper reusing `EpisodeFilenameParser` (no `os.walk`, ranking hint only).
- **Dependency** — pinned `watchdog>=2.1.0,<4.0.0` to stay compatible with the currently declared minimum Python (deferring any deliberate 3.9 bump to a maintainer decision), added to the py2exe bundle.

## The §5 direct-watch gate (validated on the reported Win 11 + Samba setup)

A **non-recursive** `WindowsApiObserver` watch placed **directly on an existing deep directory** received all remote (Pi-originated) create/rename/delete events — whereas the recursive root watch missed them. So the supplemental per-directory watch strategy is worthwhile: it catches exactly the remotely-added media a recursive root watch does not surface at depth.

## Testing

- Event application (§11): idempotent create/delete/move, subtree removal, `TV`/`TV2` boundary safety, case-insensitive key reuse, move in/out of roots — unit tested (13/13).
- Construct-under-reactor integration: start → full scan → cache populate → direct event applied → coalesced notification → clean shutdown.
- `FileMonitor` live against the Samba share: supplemental non-recursive watch delivers deep remote events on the reactor thread; budget accounting correct.
- From-source startup and a frozen py2exe build that bundles `filemonitor` + `watchdog` (incl. the Windows `read_directory_changes`/`winapi` observer); real-session behaviour smoke (join/playlist/EOF/file-switch) showed no seek/pause/playlist/protocol change.
- **Still to run with help:** the watch-budget benchmark (Test B: 1/8/16/32/64 network watches — thread count, CPU, memory, reconnect behaviour) and the broader matrix on macOS/Linux.

Happy to adjust any of the constants, naming, or the Python/watchdog version decision.

---
Changelog line: `#799 by @AlySerry0, resolving #796 and #130`


Et0h:
Cheers, I'll review this when I have the time.

Issue #804, opened 2026-08-28
Curly braces aren't escaped correctly in syncplay chat in mpv


SquirrelKiev:
**Describe the bug**
Curly braces aren't escaped correctly in syncplay chat in mpv.

<img width="549" height="129" alt="Image" src="https://github.com/user-attachments/assets/46569137-40b6-4c08-af8b-7d6c3667af7e" />

<img width="280" height="96" alt="Image" src="https://github.com/user-attachments/assets/ef489c5f-2f07-43e2-949e-096b465d15f1" />

**To Reproduce**
1. Send a { or } in syncplay chat with mpv as the video player.
2. { gets written as `{\\fscx0}  {` instead.

**Expected behavior**
I think the intention is for it to just get displayed as `{`. (there are uses for it not being escaped, e.g. formatting chat messages with {\i1}{\i0), although that's a different issue.)

**Version and platform:**
 - OS: Windows 11
 - Syncplay version and build type: Syncplay 1.7.7 development, seen on past versions too however.
 - Media player and version: mpv v0.41.0-1010-gc318236b8

**Additional context**
I've only checked mpv for this, but if some other supported video players also display chat, it might also be an issue there too.

PR #810, opened 2026-09-20
Fix Arabic, Hebrew and Persian text in mpv chat and the chat log

yuroyami:
## Problem

When a text (written in RTL language) In the mpv chat and in mpv notifications is displayed, the letters of each word show in the opposite order. The word `بنجاح` shows as `حاجنب`. Numbers next to this text also change order. The time `10:30` shows as `03:01`.

In the chat log, the `!` and `?` characters go to the wrong end of the line. An Arabic user name can push the time stamp into the middle of the line.

## Cause

On mpv: `wordwrapify_string` puts `{\fscx0}  {\fscx100}` between all characters. This marker contains two spaces. You cannot see the spaces, but the Unicode direction rules still use them. libass then shows each Arabic letter as a separate word.

On Syncplay client: the chat log has a different cause. All messages are in one Qt paragraph, and Qt gives one direction to a full paragraph.

## Changes

`syncplayintf.lua` does not add the wrap markers to a line that contains RTL letters. libass breaks such a line at the spaces. It also breaks a word that is too long for the screen.

`gui.py` and `mpv.py` put Unicode isolate marks (U+2068 and U+2069) around each chat line and each user name. You cannot see the marks. Each part of the line keeps its own direction.

The chat log removes these marks again when a user copies or drags text. The copied data stays the same as before this change, in plain text, HTML, Markdown and OpenDocument.

## What stays the same

Text without RTL letters shows exactly as before. The mpv screenshots have 0 different pixels. The layout stays left to right.

## Limits

A notification is checked as one text. If one line of it has RTL letters, all lines of that notification lose the wrap markers.

A copy from the chat log also removes U+2068 and U+2069 characters that a user typed.

## Tests

I tested on macOS with Python 3.14, PySide6 6.11.2, mpv 0.40.0 and libass 0.17.4.

For mpv, I sent chat messages and notifications through the real `syncplayintf.lua`. I compared the screenshots from before and after the change.

For the chat log, I sent messages through `MainWindow.showMessage` into a `QTextBrowser`, before and after the change.

For the copy, I compared the copied data against the same log without the marks. I used 16 selections, in light and dark mode, with Arabic text and emoji. All four formats are the same. A copy of a full log with 2,000 messages takes about 50 ms.

I did not test drag and drop, the Windows (PySide2) build or the Linux build. VLC, MPC-HC and mplayer have no code change.

## Screenshots

mpv chat, before and after:

mpv-before-after

Chat log, before and after:

chatlog-before-after

This change follows the wrap changes from #804.


yuroyami:
This clears the way for an Arabic translation I'm working on next. Right now RTL text isn't readable so the translation has nothing to land on until this merges.

